PR #72, opened 2019-11-27
feat: add verifyWithIndividualChecks

sebastianquek:
Returns an array of promises, each promise corresponds to a verification check.
The last promise resolves to the overall validity based on all the checks.

This allows us to work on the promises independently from the other verification checks.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -3,6 +3,37 @@ import { verifyHash } from "./hash/hash";
 import { verifyIssued } from "./issued/verify";
 import { verifyRevoked } from "./revoked/verify";
 import { documentToSmartContracts } from "./common/smartContract/documentToSmartContracts";
+import { OpenAttestationContract } from "./types";
+
+type ResolveType<T> = T extends (...args: any[]) => Promise<infer U> ? U : T;
+
+type VerificationChecks = [
+  ReturnType<typeof verifyHash>,
+  ReturnType<typeof verifyIssued>,
+  ReturnType<typeof verifyRevoked>
+];
+
+type VerificationChecksWithValidity = [
+  ReturnType<typeof verifyHash>,
+  ReturnType<typeof verifyIssued>,
+  ReturnType<typeof verifyRevoked>,
+  Promise<boolean>
+];
+
+const getAllChecks = (
+  document: SignedDocument,
+  smartContracts: OpenAttestationContract[]
+): VerificationChecks => [
+  verifyHash(document),
+  verifyIssued(document, smartContracts),
+  verifyRevoked(document, smartContracts)
+];
+
+const isDocumentValid = (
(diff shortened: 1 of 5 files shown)


src/index.ts
```diff
@@ -3,6 +3,37 @@ import { verifyHash } from "./hash/hash";
 import { verifyIssued } from "./issued/verify";
 import { verifyRevoked } from "./revoked/verify";
 import { documentToSmartContracts } from "./common/smartContract/documentToSmartContracts";
+import { OpenAttestationContract } from "./types";
+
+type ResolveType<T> = T extends (...args: any[]) => Promise<infer U> ? U : T;
```
Nebulis: I think this one deserve a comment to explain what we are trying (comment + example maybe)

src/verify.integration.test.ts
```diff
@@ -152,3 +152,174 @@ describe("verify(integration)", () => {
     );
   });
 });
+
+describe("verifyWithIndividualChecks(integration)", () => {
+  it("returns false if document is invalid", async () => {
```
Nebulis: I think `document is invalid` is not clear enough on the intent, From the expectation:
- document has not been issued
- hash is invalid

What are we testing in that case ?

src/verify.test.ts
```diff
@@ -61,3 +63,56 @@ describe("verify", () => {
     });
   });
 });
+
+describe("verifyWithIndividualChecks", () => {
```
Nebulis: what does this test bring compare to integration test ?

src/verify.integration.test.ts
```diff
@@ -152,3 +152,174 @@ describe("verify(integration)", () => {
...
+            address: "0x48399Fb88bcD031C556F53e93F690EEC07963Af3",
+            revoked: false
+          }
+        ]
+      },
+      valid: true
+    });
+  });
+
+  it("returns false if Ropsten token is invalid", async () => {
```
Nebulis: I dont understand the meaning of this ? what are we testing :thinking:

Nebulis: Same as above, the test description is not describing enough what we are testing
- document has not been issued
- document has been revoked

What are we testing ?

src/index.ts
sebastianquek: OK, I've added some comments about this

src/verify.test.ts
sebastianquek: The integration test ensures the checks work on an actual document with the correct network specified (also ensures the various checks work properly together).

The unit test, on the other hand, mocks what's returned by the various checks and checks whether the returned values are correct given these mocks. Also, we can mimic the way the various promises resolve at different times.

How should we better spilt the test concerns?

sebastianquek: As discussed, I've removed the unit tests that don't value add much.

PR #73, opened 2019-12-11
feat: upgrade to open attestation v3

Nebulis:
Support for open-attestation v3. Will be merged before #74 but after https://github.com/Open-Attestation/open-attestation/pull/57

Many changes concern typings only. The feature is located in `documentToSmartContracts` function. I added tests to support v3 open-attestation document.

If you want to try it locally:
- git clone git@github.com:Open-Attestation/open-attestation.git
- git checkout feat/add-open-attestation-schema
- npm link
- cd /path/to/oa-verify
- npm link @govtechsg/open-attestation


Nebulis:
upgraded while implementing verifier manager

PR #74, opened 2019-12-12
feat: implement verifier manager

Nebulis:
Quick summary

```
So the grand idea is that we will have a verificationManager that can take in verificationMethods

so each verificationMethod will have:
- a test to check if that method will run
- the verification mechanism itself

so, for the current one, we will have likely:
- OpenAttestationTamperCheck
- OpenAttestationEthereumDocumentStoreIssued
- OpenAttestationEthereumDocumentStoreRevoked
- OpenAttestationDnsTxtIdentity

it will also be loaded with:
- OpenCertsRegistryIdentity

each of these will be an object 
{
  test: (Document) => boolean,
  verify: (Document) => VerificationFragment
}

A VerificationFragement will be like: 

{
  type: "OpenAttestationDnsTxtIdentity",
  data: "mpa.gov.sg",
  message: "Error or warning message",
  status: "VALID" | "WARNING" | "ERROR"
}

The verificationManager will summarise all the Fragments and provide a summary as well (for those dev who are too lazy to go further)

The verification manager can also take in conditions to determine the summary status: ie Tamper && Issued && Revoke && (DNS || Registry), etc
```

## Implementation details
- `verificationManager` is a function that takes a list of `verifiers` and return a function. The function expect a signed document and some options as parameter. Options parameter expect the following:
  - a mandatory `network` string (for instance ropsten)
  - an optional `infuraApiKey` string
  - an optional `promisesCallback` callback function that will provide back the promises resolving to the verifications.
- There are 4 `verifiers`:
  - openAttestationTamperCheck
  - openAttestationEthereumDocumentStoreIssued
  - openAttestationEthereumDocumentStoreRevoked
  - openAttestationDnsTxtIdentity
- There is a `defaultVerificationManager` pre-built having in parameter all the verifiers available in this repository
- There are 2 validations functions: `isValid` and `isInvalid`
- A `verifier` is an object that expect 3 functions:
  - test: who must return true or false. The function must indicate whether condition are fulfilled for the verifier to run on a specific signed document
  - verify:  who must return the result of the verification as a VerificationFragment
  - skip: who must return the result of a verification when it's skipped by providing additional data on why the validation didn't run.

## TODO
- add tests


src/common/smartContract/contractInstance.ts
```diff
@@ -1,17 +1,14 @@
...
   contractAddress: Hash;
+  infuraApiKey?: string;
 }
-export const contractInstance = ({
-  abi,
-  network,
-  contractAddress
-}: ContractInstance) => {
-  const provider = new ethers.providers.InfuraProvider(network, INFURA_API_KEY);
+export const contractInstance = ({ abi, network, contractAddress, infuraApiKey }: ContractInstance) => {
```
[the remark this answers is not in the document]
Nebulis: I just added a way to provide a custom infura api key as suggested by Raymond. I'm not sure how to change this, what is your suggestion ?

Nebulis:
I've added tests for verifiers (testing the verifier only not internals)
Didn't add test for the managers => there are already integration tests
Removed apikey as discussed

src/verifiers/verificationManager.ts
```diff
@@ -0,0 +1,22 @@
...
+  options: VerificationManagerOptions
+): Promise<VerificationFragment[]> => {
+  const promises: Promise<VerificationFragment>[] = verifiers.map(verifier => {
+    if (verifier.test(document, options)) {
+      return verifier.verify(document, options);
+    }
+    return verifier.skip(document, options);
+  });
+
+  options.promisesCallback?.(promises);
```
Nebulis: test it MF

src/verify.integration.test.ts
```diff
@@ -2,324 +2,268 @@
...
             }
-          ]
+          ],
+          issuedOnAll: false
         },
-        revoked: {
-          revokedOnAny: false,
+        message: "Certificate has not been issued",
+        status: "INVALID",
+        type: "OpenAttestationEthereumDocumentStoreIssued"
```
yehjxraymond: Am thinking that the type should be succinct and general, serving more as a class of verification. We can put these as `name` of the verification method instead.

Valid class of checks: 
`INTEGRITY` class verification methods checks if the document has been tampered with.
`ISSUANCE_STATUES` class verification methods checks if the document has been issued and is in the right status (ie. not revoked)
`ISSUER_IDENTITY` class verification methods checks for the issuer's identity.

In this case the default methods will be:
OpenAttestationTamperCheck -> `INTEGRITY`
OpenAttestationEthereumDocumentStoreIssued => `ISSUANCE_STATUES`
OpenAttestationEthereumDocumentStoreRevoked => `ISSUANCE_STATUES`
OpenAttestationDnsTxtIdentity => `ISSUER_IDENTITY`

Possible extension will look like: 
OpenAttestationEthereumTokenRegistryIssued => `ISSUANCE_STATUES`
OpenCertsIdentityRegistry  => `ISSUER_IDENTITY`
SomeCountryIdentityRegistry  => `ISSUER_IDENTITY`
CustomExpiryCheck => `ISSUANCE_STATUES`

Using a known class, it can allow client application to check what type of checks has been performed it they like.

src/verify.integration.test.ts
```diff
@@ -2,324 +2,268 @@
...
-          }
-        ]
+        status: "VALID",
+        type: "OpenAttestationEthereumDocumentStoreRevoked"
       },
-      valid: true
-    });
+      {
+        message: 'Document issuers doesn\'t have "documentStore" or "token" property',
+        status: "SKIPPED",
```
yehjxraymond: Just a question... 

If I have the `OpenCertsRegistry` extension:
1. This method will show skipped
2. OpenCertsRegistry method will show valid

How should I go about handling it in the method `isValid`?

src/index.ts
```diff
@@ -1,96 +1,32 @@
...
-    .then(() => true)
-    .catch(() => false);
-
-  return [hash, issued, revoked, valid];
+const defaultVerificationManager = verificationManager(defaultVerifiers);
+const isValid = (verificationFragments: VerificationFragment[]) =>
+  verificationFragments.every(
+    verificationFragment => verificationFragment.status === "VALID" || verificationFragment.status === "SKIPPED"
+  );
+const isInvalid = (verificationFragments: VerificationFragment[]) => !isValid(verificationFragments);
```
Nebulis: add tests MF

src/index.ts
```diff
@@ -1,96 +1,32 @@
...
-  ReturnType<typeof verifyIssued>,
-  ReturnType<typeof verifyRevoked>
-];
-
-type VerificationChecksWithValidity = [
-  ReturnType<typeof verifyHash>,
-  ReturnType<typeof verifyIssued>,
-  ReturnType<typeof verifyRevoked>,
-  Promise<boolean>
+import { verificationManager } from "./verifiers/verificationManager";
```
yehjxraymond:
```suggestion
import { verificationBuilder } from "./verifiers/verificationBuilder";
```

src/index.ts
```diff
@@ -1,96 +1,32 @@
...
-    ),
-    new Promise(async (resolve, reject) =>
-      (await revoked).revokedOnAny ? reject() : resolve()
-    )
-  ])
-    .then(() => true)
-    .catch(() => false);
-
-  return [hash, issued, revoked, valid];
+const defaultVerificationManager = verificationManager(defaultVerifiers);
```
yehjxraymond:
```suggestion
const verify = verificationBuilder(defaultVerifiers);
```

src/index.ts
```diff
@@ -1,96 +1,32 @@
...
+  verificationManager,
+  isValid,
+  isInvalid,
+  openAttestationEthereumDocumentStoreIssued,
+  openAttestationDnsTxtIdentity,
+  openAttestationTamperCheck,
+  openAttestationEthereumDocumentStoreRevoked,
+  defaultVerifiers,
+  defaultVerificationManager
 };
```
yehjxraymond:
```
export {
  verify,
  verificationBuilder,
  isValid,
  isInvalid,
  defaultVerifiers,
  verifiers
}
```

src/index.ts
```diff
@@ -1,96 +1,32 @@
...
-    new Promise(async (resolve, reject) =>
-      (await revoked).revokedOnAny ? reject() : resolve()
-    )
-  ])
-    .then(() => true)
-    .catch(() => false);
-
-  return [hash, issued, revoked, valid];
+const defaultVerificationManager = verificationManager(defaultVerifiers);
+const isValid = (verificationFragments: VerificationFragment[]) =>
```
yehjxraymond: isValid(document, requiredChecksClasses)

usage example
isValid(document, ["ISSUER_IDENTITY", "ISSUANCE_STATUS", "INTEGRITY"])

yehjxraymond: Basically isValid checks the fragments that EVERY required checks classes to have at EVERY test either "VALID" or "SKIPPED" and has at least one "VALID".

For example for a documentStore
issued: valid
revoked: invalid
will result in "invalid"

In another case: 
issued: valid
revoked: skipped
will result in "valid"

src/verify.integration.test.ts
Nebulis: done

src/index.ts
Nebulis: done

src/index.ts
Nebulis: done

src/index.ts
Nebulis: done

src/verifiers/verificationManager.ts
Nebulis: will skip test for this one, I'm a bit lazy sorry =). Blame me if there is a bug on this one, I will fix and write tests

src/types/core.ts
```diff
@@ -0,0 +1,74 @@
...
+}
+export type Hash = string;
+
+export interface OpenAttestationContract {
+  address: Hash;
+  type: v3.Method;
+  instance: Contract;
+}
+
+export const isWrappedV3Document = (document: any): document is WrappedDocument<v3.OpenAttestationDocument> => {
```
yehjxraymond: Is this weird to be in the `types` folder?

src/types/core.ts
```diff
@@ -0,0 +1,74 @@
...
+export type VerificationFragmentType = "DOCUMENT_INTEGRITY" | "DOCUMENT_STATUS" | "ISSUER_IDENTITY";
+export type VerificationFragmentStatus = "ERROR" | "VALID" | "INVALID" | "SKIPPED";
+
+/**
+ * A verifier is an object whose goal is to perform specific validation on a signed document. It exposes
+ * - a *test* function, who must return true or false. The function must indicate whether condition are fulfilled for the verifier to run on a specific signed document
+ * - a *verify* function, who must return the result of the verification as a {@link VerificationFragment}
+ * - a *skip* function, who must return the result of a verification when it's skipped by providing additional data on why the validation didn't run.
+ */
+interface SkippedVerificationFragment {
```
yehjxraymond: Why does this warrant an interface of it's own? Also does "SKIPPED" conflict with enum "SKIPPED"?

Was wondering if it must be used, can we extend `VerificationFragment`

src/hash/hash.ts
```diff
@@ -1,8 +1,5 @@
-import {
-  verifySignature,
-  SchematisedDocument
-} from "@govtechsg/open-attestation";
+import { verifySignature, WrappedDocument } from "@govtechsg/open-attestation";
```
yehjxraymond: Seemed like this is no longer needed in this folder. Might want to put it together with the `openAttestationHash` function.

yehjxraymond: On that note, perhaps we want to think about how to have these verifiers provide utility functions as well. For instance the identity verifiers can provide a function that returns the resolved identities. Can create an issue to track it instead of doing it in this PR.

src/issued/contractInterface.ts
```diff
@@ -1,31 +1,22 @@
 import { constants } from "ethers";
-import { Hash, OpenAttestationContract } from "../types";
-import { TYPES } from "../common/smartContract/constants";
+import { v3 } from "@govtechsg/open-attestation";
```
yehjxraymond: Shall we remove this from the `issued` folder and put them with the verifiers?

src/revoked/contractInterface.integration.test.ts
```diff
@@ -1,114 +1,86 @@
 import { constants } from "ethers";
 
+import { isRevoked, isRevokedOnDocumentStore, isRevokedOnTokenRegistry } from "./contractInterface";
```
yehjxraymond: Same comment to put with the verifiers

src/verifiers/openAttestationEthereumTokenRegistryMinted.ts
```diff
@@ -0,0 +1,46 @@
...
+      message: `Document issuers doesn't have "tokenRegistry" property or ${v3.Method.TokenRegistry} method`
+    });
+  },
+  test: document => {
+    if (isWrappedV3Document(document)) {
+      const documentData = getData(document);
+      return documentData.proof.method === v3.Method.TokenRegistry;
+    }
+    const documentData = getData(document);
+    return documentData.issuers.every(issuer => "tokenRegistry" in issuer);
```
yehjxraymond: For token registry, we should check of `any`/`some` instead of `every`.

The reason is that if it's using tokenRegistry it **MUST** be only one issuer, cannot mix it around with documentstore or other tokenregistry. 

In the verify function we can throw `INVALID` if it has been mixed

src/verifiers/openAttestationEthereumDocumentStoreIssued.ts
```diff
@@ -0,0 +1,46 @@
...
+      message: `Document issuers doesn't have "documentStore" or "certificateStore" property or ${v3.Method.DocumentStore} method`
+    });
+  },
+  test: document => {
+    if (isWrappedV3Document(document)) {
+      const documentData = getData(document);
+      return documentData.proof.method === v3.Method.DocumentStore;
+    }
+    const documentData = getData(document);
+    return documentData.issuers.every(issuer => "documentStore" in issuer || "certificateStore" in issuer);
```
yehjxraymond: Should we test using `some` instead? 
This will make the verify run for documents that mixes around documentStore and other methods, and that's where this function should throw?

src/types/core.ts
Nebulis: ~ slightly didnt want to create a file for that, can refactor later

src/types/core.ts
Nebulis: SKIPPED is not an enum but a literal => no conflicts

It could extends `VerificationFragment` at least dont need to add `name` and `type` but it wont provide anything else

=> will use extends (done)

src/hash/hash.ts
Nebulis: removed and added issue

src/issued/contractInterface.ts
Nebulis: moved under verifiers/documentStoreIssued

src/revoked/contractInterface.integration.test.ts
Nebulis: moved under verifiers/documentStoreRevoked

src/verifiers/openAttestationEthereumTokenRegistryMinted.ts
Nebulis: done, returning error if more than one token registry or mix of methods

src/verifiers/openAttestationEthereumDocumentStoreIssued.ts
Nebulis: done, returning error if one of the issuer is invalid (i.e. doesnt have document store or certificate store)

PR #77, opened 2019-12-31
fix: dns verifier

Nebulis:
- only verify if the issuer is using DNS-TXT (dont throw error if he's not)
- added resolution status for every issuer, even if it has been skipped for one person
- forward the resolution status in the data object

### DIFF
--- a/src/verifiers/openAttestationDnsTxt.ts
+++ b/src/verifiers/openAttestationDnsTxt.ts
@@ -14,6 +14,7 @@ type Identity =
       value: string;
     };
 // Resolve identity of an issuer, currently supporting only DNS-TXT
+// DNS-TXT is explained => https://github.com/Open-Attestation/adr/blob/master/decentralized_identity_proof_DNS-TXT.md
 const resolveIssuerIdentity = async (
   issuer: v2.Issuer | v3.Issuer,
   smartContractAddress: string,
--- a/src/verifiers/openAttestationDnsTxt.ts
+++ b/src/verifiers/openAttestationDnsTxt.ts
@@ -106,29 +106,31 @@ export const openAttestationDnsTxt: Verifier<
           data: identities,
           status: "VALID"
         };
-      }
-      const documentData = getData(document);
-      const identity = await resolveIssuerIdentity(documentData.issuer, documentData.proof.value, options);
-      if (identity.status === "INVALID") {
+      } else {
+        // v3 document
+        const documentData = getData(document);
+        const identity = await resolveIssuerIdentity(documentData.issuer, documentData.proof.value, options);
+        if (identity.status === "INVALID") {
+          return {
+            name,
+            type,
+            data: {
+              type: documentData.issuer.identityProof.type,
+              location: documentData.issuer.identityProof.location,
+              value: documentData.proof.value
+            },
+            message: "Certificate issuer identity is invalid",
+            status: "INVALID"
+          };
+        }
+
         return {
           name,
           type,
-          data: {
-            type: documentData.issuer.identityProof.type,
-            location: documentData.issuer.identityProof.location,
-            value: documentData.proof.value
-          },
-          message: "Certificate issuer identity is invalid",
-          status: "INVALID"
+          data: identity,
+          status: "VALID"
         };
       }
-
-      return {
-        name,
-        type,
-        data: identity,
-        status: "VALID"
-      };
     } catch (e) {
       return {
         name,
--- a/src/verifiers/openAttestationDnsTxt.ts
+++ b/src/verifiers/openAttestationDnsTxt.ts
@@ -89,14 +89,14 @@ export const openAttestationDnsTxt: Verifier<
 
         const invalidIdentity = identities.findIndex(identity => identity.status === "INVALID");
         if (invalidIdentity !== -1) {
-          const value =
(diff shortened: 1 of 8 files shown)


src/verifiers/openAttestationDnsTxt.ts
```diff
@@ -103,7 +109,7 @@ export const openAttestationDnsTxt: Verifier<
       }
       const documentData = getData(document);
```
rjchow: either a comment or an explicit `else if` here would be nice because i was confused about where v3 was being handled when i saw v2 was being handled above

src/verifiers/openAttestationDnsTxt.ts
```diff
@@ -73,24 +75,28 @@ export const openAttestationDnsTxt: Verifier<
       if (isWrappedV2Document(document)) {
         const documentData = getData(document);
         const identities = await Promise.all(
-          // we expect the test function to prevent this issue => smart contract address MUST be populated
-          // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-          documentData.issuers.map(issuer => resolveIssuerIdentity(issuer, getSmartContractAddress(issuer)!, options))
+          documentData.issuers.map(issuer => {
```
rjchow: what happens if more than 1 issuer fails?

src/verifiers/openAttestationDnsTxt.v2.test.ts
```diff
@@ -148,12 +305,66 @@ describe("OpenAttestationDnsTxt v2 document", () => {
...
+            },
+            {
+              status: "SKIPPED"
+            }
+          ],
+          status: "VALID"
+        }
+      ]);
+    });
+    it("should return skipped fragment if no issuer has a tokenRegistry or documentStore", async () => {
```
rjchow: i assume there's somewhere that fails when the document doesn't have a valid issuer check?

src/verifiers/openAttestationDnsTxt.ts
```diff
@@ -3,18 +3,15 @@ import { getDocumentStoreRecords } from "@govtechsg/dnsprove";
...
     }
   | {
-      identified: false;
-      smartContract: string;
-      error?: string | Error;
+      status: "INVALID";
+      value: string;
     };
 // Resolve identity of an issuer, currently supporting only DNS-TXT
 const resolveIssuerIdentity = async (
```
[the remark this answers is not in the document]
Nebulis: ok to link to ADR ?

src/verifiers/openAttestationDnsTxt.ts
Nebulis: we are all doomed :see_no_evil:

src/verifiers/openAttestationDnsTxt.v2.test.ts
Nebulis: 1. https://github.com/Open-Attestation/oa-verify/blob/fix/dns-verifier/src/verifiers/openAttestationDnsTxt.v2.test.ts#L63
2. https://github.com/Open-Attestation/oa-verify/blob/fix/dns-verifier/src/verifiers/openAttestationDnsTxt.v2.test.ts#L282

src/verifiers/openAttestationDnsTxt.ts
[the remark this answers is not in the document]
rjchow: ok

src/verifiers/openAttestationDnsTxt.ts
Nebulis: remove no-else-return eslint rules, added a comment and else block :)

src/verifiers/openAttestationDnsTxt.ts
```diff
@@ -73,24 +75,28 @@ export const openAttestationDnsTxt: Verifier<
...
+            return {
+              status: "SKIPPED"
+            };
+          })
         );
 
-        const invalidIdentity = identities.findIndex(identity => !identity.identified);
+        const invalidIdentity = identities.findIndex(identity => identity.status === "INVALID");
         if (invalidIdentity !== -1) {
+          const value =
```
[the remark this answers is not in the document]
Nebulis: renamed to smartContractAddress

src/verifiers/openAttestationDnsTxt.ts
rjchow: or you could do else if (test for v3) { }
else { throw new Error("wtf you ain't supposed to be here") }

src/verifiers/openAttestationDnsTxt.ts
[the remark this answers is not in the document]
Nebulis: added link to ADR

src/verifiers/openAttestationDnsTxt.ts
Nebulis: Nah, the check is just to get correct typings and autocomplete, otherwise we wouldn't do it. in the `else` branch`, type is correct, sad if you did something weird it will probably throw and be caught by the try/catch block

src/verifiers/openAttestationDnsTxt.ts
Nebulis: Created issue to handle that case https://github.com/Open-Attestation/oa-verify/issues/78

PR #80, opened 2020-01-09
feat: 🎸 Add verifier for w3c-did identityProof

AdamJLemmon:
- Support for universal resolver for broad did resolution. 
- Initial did-auth support for ethr did-method utilizing comparison against tx.origin of document issuance tx.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,8 +22,10 @@
   "license": "GPL-3.0",
   "dependencies": {
     "@govtechsg/dnsprove": "^2.0.5",
-    "@govtechsg/open-attestation": "3.0.4",
-    "ethers": "^4.0.40"
+    "@govtechsg/open-attestation": "3.1.0",
+    "@types/request-promise-native": "^1.0.17",
+    "ethers": "^4.0.40",
+    "request-promise-native": "^1.0.8"
   },
   "devDependencies": {
     "@commitlint/cli": "8.2.0",
--- a/src/types/w3c-did.ts
+++ b/src/types/w3c-did.ts
@@ -1,0 +1,63 @@
+export interface Identity {
+  identified: true;
+  ethereumAddress: string;
+  smartContract: string;
+}
+
+export interface DIDDocument {
+  "@context": "https://w3id.org/did/v1";
+  id: string;
+  publicKey: PublicKey[];
+  authentication?: Authentication[];
+  uportProfile?: any;
+  service?: ServiceEndpoint[];
+  created?: string;
+  updated?: string;
+  proof?: LinkedDataProof;
+}
+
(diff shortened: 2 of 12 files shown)


src/types/w3c-did.ts
```diff
@@ -0,0 +1,63 @@
+export interface Identity {
+  identified: true;
+  ethereumAddress: string;
+  smartContract: string;
+}
+
+export interface DIDDocument {
```
rjchow: I think you can use some of the types/methods from here https://github.com/decentralized-identity/did-common-typescript/blob/master/lib/IDidDocument.ts

src/verifiers/openAttestationW3CDID.ts
```diff
@@ -0,0 +1,70 @@
...
+ *  - Determine if a standard format to be used for the `data` field within returned fragments.
+ *    Currently the didDoc is returned as the `data` field.
+ */
+export const openAttestationW3CDID: Verifier<WrappedDocument<v3.OpenAttestationDocument>> = {
+  skip: () => {
+    return Promise.resolve({
+      status: "SKIPPED",
+      type,
+      name,
+      message: `Document issuer doesn't doesn't use ${v3.IdentityProofType.W3CDid} type`
```
rjchow: doesn't doesnt :p

package.json
```diff
@@ -22,8 +22,10 @@
   "license": "GPL-3.0",
   "dependencies": {
     "@govtechsg/dnsprove": "^2.0.5",
-    "@govtechsg/open-attestation": "3.0.4",
-    "ethers": "^4.0.40"
+    "@govtechsg/open-attestation": "3.1.0",
+    "@types/request-promise-native": "^1.0.17",
+    "ethers": "^4.0.40",
+    "request-promise-native": "^1.0.8"
```
rjchow: we usually use axios with async/await

src/verifiers/openAttestationW3CDID.ts
AdamJLemmon: :+1:

src/types/w3c-did.ts
AdamJLemmon: I actually pulled direct from the did-resolver here: https://github.com/decentralized-identity/did-resolver/blob/02bdaf1687151bb934b10093042e576ed54b229c/src/resolver.ts#L15

rjchow: would it make sense to then install it as a dependency and use the exported types?

AdamJLemmon: That feels like a good solution.  Then also as types inevitably evolve we can just bump version of the package instead of having to replicate changes manually... I like it.

PR #82, opened 2020-01-09
fix: handle certificate store with dns txt

Nebulis:
According to @rjchow we dont need that, opening for a while :D


Nebulis:
Never trust @rjchow.
Never.

PR #91, opened 2020-04-26
feat: 🎸 verify signed proof blocks

AdamJLemmon:
Allow the documents to be signed diretly and a proof block appended. Extend the verifier to use this proof block if it exists to verify if the document has been issued instead of querying the Ethereum based document store.  Allows real-time issuance with no dependence on the blockchain 
 (transaction inclusion) and thus no associated tx costs either.  We also have a reference implementation for the signature generation and appending the proof we could share.


src/verifiers/documentStoreIssued/openAttestationDocumentStoreIssued.ts
```diff
@@ -0,0 +1,24 @@
+import { v2, v3, WrappedDocument } from "@govtechsg/open-attestation";
```
Nebulis, line 1: Any reason to add this ? I feel this is unnecessary :thinking: 
- can add openAttestationW3CDIDProof in the list of verifiers (index)
- `openAttestationDocumentStoreIssued` wont run if there is no document store (thanks to skip)
- `openAttestationW3CDIDProof` wont run if there is no proof (thanks to skip)

that way we have a result for each verifiers

src/verifiers/w3c-did/resolveDID.ts
```diff
@@ -0,0 +1,19 @@
...
+ * @todo
+ *  - Consider... currently reliant on external service @ `BASE_RESOLVER_URL` staying online / being accessible from wherever
+ *    the verifier is run. Could look to host our own univeral resolver instance as per:
+ *    https://github.com/decentralized-identity/universal-resolver#quick-start
+ */
+export const resolveDID = async (did: string): Promise<DIDDocument> => {
+  const uri = `${BASE_RESOLVER_URL}/${RESOLVER_VERSION}/${RESOLVER_PATH}/${did}`;
+  const resolvedDID: any = await axios.get(uri);
+  const { didDocument } = resolvedDID.data;
+  if (!didDocument) throw new Error(`DIDDoc could not be retireved for: ${did}`);
```
Nebulis, line 17: typo => retrieved

test/fixtures/v2/documentRopstenInvalidWithDIDSignedProofBlockProperHash.ts
```diff
@@ -0,0 +1,55 @@
+import { v2, WrappedDocumentWithProof } from "@govtechsg/open-attestation";
```
Nebulis, line 1: what is invalid ?

test/fixtures/v2/documentRopstenValidWithDIDSignedProofBlockProperHash.ts
```diff
@@ -0,0 +1,55 @@
+import { v2, WrappedDocumentWithProof } from "@govtechsg/open-attestation";
```
Nebulis, line 1: Questions regarding the 3 test files added
1. how did you add the proof (manually ? ur own tool ? I think if we want this we need to add it into open-attestation)
1. Did you sign the whole content ?
1. Any reason you kept the certificate store ? (other than open-attestation require one but then it's not good and we need to update oa)

src/config.ts
```diff
@@ -1 +1,11 @@
 export const INFURA_API_KEY = "92c9a51428b946c1b8c1ac5a237616e4";
+
+// DID Universal Resolver parameters
+export const BASE_RESOLVER_URL = "https://uniresolver.io";
+export const RESOLVER_VERSION = "1.0";
+export const RESOLVER_PATH = "identifiers";
+
+// w3c-dids
+export const ETHR_DID_METHOD = "ethr";
+export const SUPPORTED_DID_AUTH = ["Secp256k1SignatureAuthentication2018", "Secp256k1SignatureAuthentication2019"];
```
Nebulis, line 10: no 2020, no 2017, etc. ?

sorry I'm noob on this, what's the diff between 2018/2019 ? just the name or there is more ?

src/verifiers/w3c-did/openAttestationW3CDIDProof.ts
```diff
@@ -0,0 +1,82 @@
...
+  verify: async document => {
+    try {
+      const { proof, signature } = document;
+      const didDoc = await resolveDID(proof.verificationMethod);
+
+      // Confirm this type of proof is currently supported
+      if (!SUPPORTED_PROOF_TYPES.includes(proof.type)) throw new Error(`Proof type: ${proof.type} is not supported.`);
+
+      // If the didDoc has an auth method we currently support
+      const supportedAuth = didDoc?.authentication?.filter((auth: Authentication) =>
```
Nebulis, line 37: can use find instead of filter

src/verifiers/w3c-did/openAttestationW3CDIDProof.ts
```diff
@@ -0,0 +1,82 @@
...
+      }
+    });
+  },
+  test: document => {
+    return Object.keys(document).includes("proof");
+  },
+  verify: async document => {
+    try {
+      const { proof, signature } = document;
+      const didDoc = await resolveDID(proof.verificationMethod);
```
Nebulis, line 31: rename to didDocument ? :flushed:

src/verifiers/w3c-did/openAttestationW3CDIDProof.ts
```diff
@@ -0,0 +1,82 @@
...
+      if (!SUPPORTED_PROOF_TYPES.includes(proof.type)) throw new Error(`Proof type: ${proof.type} is not supported.`);
+
+      // If the didDoc has an auth method we currently support
+      const supportedAuth = didDoc?.authentication?.filter((auth: Authentication) =>
+        SUPPORTED_DID_AUTH.includes(auth.type)
+      );
+      if (!supportedAuth?.length) throw new Error("Issuer DID cannot be authenticated, no supported auth in didDoc.");
+
+      // Get the correct pub key from the ver method
+      const pubKey: any = didDoc.publicKey.find(key => key.id === proof.verificationMethod);
```
Nebulis, line 43: why typing as any ? didDoc is correctly type (from the review)
and we probably what to return invalid/error inc ase the pubKey is not found ?

src/verifiers/w3c-did/openAttestationW3CDIDProof.ts
```diff
@@ -0,0 +1,82 @@
...
+      // The exact verification may vary based on did-method (did:<method>:<namespace uuid>)
+      const didMethod = proof.verificationMethod.split(":")[1];
+      let proofValid = false;
+      if (didMethod === ETHR_DID_METHOD) {
+        // Existing targetHash is being signed
+        const msg = signature.targetHash;
+        const { ethereumAddress } = pubKey;
+        const recoverAddress = ethers.utils.verifyMessage(msg, proof.signature);
+        proofValid = recoverAddress?.toLowerCase() === ethereumAddress?.toLowerCase();
```
Nebulis, line 52: I would expect recoverAddress and ethereumAddress to not be optional here :thinking:

test/fixtures/v2/documentRopstenInvalidWithDIDSignedProofBlockProperHash.ts
AdamJLemmon: The `signature` is invalid. Changed the first 3 chars of the `signature` to `123` to be used for failing cases in the integration tests.

test/fixtures/v2/documentRopstenValidWithDIDSignedProofBlockProperHash.ts
AdamJLemmon: 1. Yes we signed using our own tooling then copied over.  Happy to pull that code into an appropriate spot in OA as you see fit.  Perhaps worth a chat about this.

2. You are right! We just signed the `targetHash` to reduce complexity of computing our own hash of the content.

3. Mainly so I could use the same `signature` block so the hash tests would pass during integration testing and to conform yes.  Also the certificate store can still be used for revocation registry in the short term if desired.

src/config.ts
AdamJLemmon: no 2020 as yet.  I believe there were slight differences in 2019 as drafted here: (https://w3c-ccg.github.io/lds-ecdsa-secp256k1-2019/).  I am not certain how versioning occurs / naming though.

test/fixtures/v2/documentRopstenInvalidWithDIDSignedProofBlockProperHash.ts
Nebulis: can you add a comment somewhere in this file ?

src/verifiers/w3c-did/openAttestationW3CDIDProof.ts
AdamJLemmon: updated `didDoc` and `pubKey` :)

src/verifiers/w3c-did/openAttestationW3CDIDProof.ts
AdamJLemmon: Correct re `recoveredAddress`, updated.  For `ethereumAddress` it is optional on the `PublicKey` type because there are other types of keys that do not have Ethereum Addresses, ie. commonly ed25519 keys are used.  I've updated to check for presence of `ethereumAddress` first now once we know we are using the `ethr` did-method.

test/fixtures/v2/documentRopstenInvalidWithDIDSignedProofBlockProperHash.ts
AdamJLemmon: yessir added a comment above where the document is defined.
```
/**
 * This document is invalid as the signature within the proof block has been tampered with.
 * The first 3 characters have been changed to 123. To be used for integration testing.
 */
```

src/verifiers/documentStoreIssued/openAttestationDocumentStoreIssued.ts
AdamJLemmon: My thought was that having both of those verifiers felt redundant as they are serving the same purpose and we only need one to be successful.  Also there may be cases where a document store and proof block are both included but the document was not issued to the document store. For example when the document stored is used as just the revocation registry for a specific issuer.  This would then fail using the `openAttestationEthereumDocumentStoreIssued` verifier but is in fact valid due to the signed proof block.

package.json
```diff
@@ -23,6 +23,7 @@
   "dependencies": {
     "@govtechsg/dnsprove": "^2.0.5",
     "@govtechsg/open-attestation": "3.1.0",
+    "axios": "^0.19.2",
```
Nebulis, line 26: to remove ?

src/verifiers/w3c-did/openAttestationW3CDIDProof.ts
```diff
@@ -0,0 +1,67 @@
+import * as ethers from "ethers";
```
Nebulis, line 1: The name of the file doesnt match anymore ^^

test/fixtures/v2/documentMainnetValidWithDIDSignedProofBlockProperHash.ts
```diff
@@ -0,0 +1,55 @@
+import { v2, WrappedDocument } from "@govtechsg/open-attestation";
```
Nebulis, line 1: need to rename the fixture too =)

test/fixtures/v2/documentRopstenInvalidProofSignature.ts
```diff
@@ -0,0 +1,51 @@
+import { v2, WrappedDocument } from "@govtechsg/open-attestation";
```
Nebulis, line 1: can you add a comment to explain why the signature not good or add a more obvious invalid signature ?

test/fixtures/v2/documentRopstenInvalidWithDIDSignedProofBlockProperHash.ts
```diff
@@ -0,0 +1,59 @@
+import { v2, WrappedDocument } from "@govtechsg/open-attestation";
```
Nebulis, line 1: change name too =)

test/fixtures/v2/documentRopstenInvalidWithDIDSignedProofBlockProperHash.ts
```diff
@@ -0,0 +1,59 @@
...
+
```
Nebulis, line 2: is this file still needed ? I see verification method still using did

test/fixtures/v2/documentRopstenValidWithDIDSignedProofBlock.ts
```diff
@@ -0,0 +1,51 @@
+import { v2, WrappedDocument } from "@govtechsg/open-attestation";
```
Nebulis, line 1: rename file

test/fixtures/v2/documentRopstenValidWithDIDSignedProofBlock.ts
```diff
@@ -0,0 +1,51 @@
...
+  schema: "opencerts/1.4",
+  data: {
+    id: "ab89a9ae-954f-4d28-8b48-2a534d3a3d60:string:2018-SAF-01",
+    $template: "2d00853b-43ae-4bc9-82a2-c614ec0fca49:string:SG-GOVTECH-OPENCERTS",
+    name: "9a4499fa-8f68-43d8-b42b-57dc365ab249:string:Certified OpenCerts Associate",
+    issuedOn: "60bf1feb-e373-4757-8a5d-cf485199bf7a:string:2018-11-30T15:00:00+08:00",
+    issuers: [
+      {
+        name: "a8d46c32-2e35-4f4e-99b1-f8a5acb04180:string:GovTech",
+        certificateStore: "9f3ffc2c-2e06-4a9d-a762-8449aec4ca9e:string:0x007d40224f6562461633ccfbaffd359ebb2fc9ba"
```
Nebulis, line 23: let's remove the need for certificate store

src/verifiers/signedProof/openAttestationSignedProof.test.ts
```diff
@@ -0,0 +1,58 @@
+import { openAttestationSignedProof } from "./openAttestationSignedProof";
```
Nebulis, line 1: Could you slightly change the way of testing like https://github.com/Open-Attestation/oa-verify/blob/master/src/verifiers/dnsText/openAttestationDnsTxt.v2.test.ts

1. build a verifier with this verification method
1. run the verifier

=> there is no need to call .test or .verify method of the verification method, just call the verifier and make sure the output is as expected
=> let's keep all the current use cases

AdamJLemmon: much nicer!

src/verifiers/signedProof/openAttestationSignedProof.test.ts
```diff
@@ -0,0 +1,58 @@
+import { verificationBuilder } from "../verificationBuilder";
+import { openAttestationSignedProof } from "./openAttestationSignedProof";
+import { documentSignedProofValid } from "../../../test/fixtures/v2/documentSignedProofValid";
+import { documentSignedProofInvalidProofType } from "../../../test/fixtures/v2/documentSignedProofInvalidProofType";
+import { documentSignedProofInvalidSignature } from "../../../test/fixtures/v2/documentSignedProofInvalidSignature";
+const verify = verificationBuilder([openAttestationSignedProof]);
+
+describe("OpenAttestationSignedProof", () => {
+  describe("v2", () => {
+    it("should return a valid fragment when document has valid signed proof", async () => {
```
Nebulis, line 10: COuld you add the test with a document that will be skipped ?

PR #95, opened 2020-06-19
chore: add tests for rinkeby network

Nebulis:


src/verify.v2.integration.test.ts
```diff
@@ -741,4 +745,184 @@ describe("verify(integration)", () => {
...
+        type: "ISSUER_IDENTITY",
+        data: [
+          {
+            status: "INVALID",
+            location: "example.openattestation.com",
+            value: "0x718B518565B81097b185661EBba3966Ff32A0039",
+          },
+        ],
+        reason: {
+          code: 1,
```
rjchow, line 820: why is this invalid?

Nebulis: I dunno but it suddenly became valid so I changed the output

PR #110, opened 2020-08-18
feat: new error handling

gjj:
### Purpose

This MR attempts to fix the issues highlighted in https://github.com/OpenCerts/opencerts-website/issues/517:

- Cloudflare Ethereum Gateway's random 502 Bad Gateway errors (something we can't control)
- Infura and Cloudflare's 429 rate limit error, caused by the user while spamming the verifier

For reference,

> A 4xx code indicates an error caused by the user, whereas 5xx codes tell the client that they did everything correctly and it's the server itself who caused the problem.

In addition to the above, we also introduced error handling for `INVALID_ARGUMENT` type of errors.

### Decision

We decided to check Ethers error codes instead of its error messages for the two new error checks introduced in this MR.

This means that:

- `SERVER_ERROR` encompasses `{ "bad response", "failed response", "missing response" }`
  - Instead of checking for Ethers error messages (e.g. `bad response`, `missing response`, etc.), we now group them using [Ethers error codes](https://docs.ethers.io/v5/api/utils/logger/#errors) wherever possible
  - Rationale is that we don't need a fine level of granularity at this point in time (i.e. we don't need to know that it's `4xx` vs `5xx`
- `INVALID_ARGUMENT` encompasses a few funky errors that we didn't catch before like `{ "incorrect data length", "hex is of odd-length" }` as explained below:
  - `61dc9186345e05cc2ae53dc72af880a3b66e2fa7983feaa6254d1518540de5` is even length, but is not 64 characters (or 32 bytes) as required of merkle roots - Ethers returns error message `incorrect data length`
  - `61dc9186345e05cc2ae53dc72af880a3b66e2fa7983feaa6254d1518540de50` is not of valid length - Ethers returns error message `hex is of odd-length`
  - `61dc9186345e05cc2ae53dc72af880a3b66e2fa7983feaa6254d1518540de50a` is OK, leaving this here for an example of a valid merkleRoot to provide context for the two errors above

This MR also includes the testcases for:

- [x] server error
- [x] invalid argument

Finally, moved `common/smartContract/documentStoreErrors.ts` → `src/verifiers/documentStoreStatus/errors.ts` since Token Registry follows the same structure.

### For future ref

HTTP-related errors from Ethers: https://github.com/ethers-io/ethers.js/blob/9640e864a68b4a9e84e820f0ceaf1eb56c66715f/packages/web/src.ts/index.ts

`Bytes`-related errors from Ethers: https://github.com/ethers-io/ethers.js/blob/a78ca7eb8d841f8dbbdda13ef8b128a8817d10a4/packages/bytes/src.ts/index.ts

tl;dr there are just simply too many error messages to handle, so handling error codes (wherever possible) is a better choice in this case


src/verifiers/documentStoreStatus/errors.ts
```diff
@@ -55,7 +70,10 @@ export const getErrorReason = (error: EthersError, address: string): Reason | nu
     (reason.toLowerCase() === "invalid address".toLowerCase() && error.code === errors.INVALID_ARGUMENT)
   ) {
     return contractAddressInvalid(address);
+  } else if (reason.toLowerCase() === "bad response".toLowerCase()) {
```
gjj, line 73: There's the `errors.SERVER_ERROR` from Ethers, should I actually `&& error.code === errors.SERVER_ERROR`?

package.json
```diff
@@ -29,6 +29,7 @@
     "@govtechsg/token-registry": "^2.0.0",
     "debug": "^4.1.1",
     "ethers": "^5.0.4",
+    "msw": "^0.20.5",
```
Nebulis, line 32: dev dependency ? :)

src/verify.v2.integration.test.ts
```diff
@@ -820,4 +824,513 @@ describe("verify(integration)", () => {
...
+          ],
+          "name": "OpenAttestationDnsTxt",
+          "status": "VALID",
+          "type": "ISSUER_IDENTITY",
+        },
+      ]
+    `);
+    expect(isValid(results)).toStrictEqual(false);
+    // Ethers would return INVALID_ARGUMENT, as merkle root is odd-length which we tampered it by removing the last char
+    expect(isValid(results, ["DOCUMENT_STATUS", "DOCUMENT_INTEGRITY"])).toStrictEqual(false);
```
Nebulis, line 907: let's split this:
```
    expect(isValid(results, ["DOCUMENT_STATUS"])).toStrictEqual(false);
    expect(isValid(results, ["DOCUMENT_INTEGRITY"])).toStrictEqual(false);
```

If one is invalid and not the other we wont catch it

Nebulis: same for the others

Nebulis (approved):

PR #112, opened 2020-08-31
chore: scaffold for new verifier & refactor verifiers

yehjxraymond:


src/verifiers/documentStatus/didSignedDocumentStatus/didSignedDocumentStatus.test.ts
```diff
@@ -0,0 +1,33 @@
...
+  it("should return skip message", () => {});
+});
+
+describe("test", () => {
+  describe("v2", () => {
+    it("should return false for documents not signed by DID", () => {});
+    it("should return true for documents where any issuer is using the `DID` identity proof", () => {});
+    it("should return true for documents where any issuer is using the `DNS-DID` identity proof", () => {});
+  });
+  describe("v3", () => {
```
Nebulis, line 13: I would suggest to keep tests for v3 skipped (but keep the description) we will implement once v3 is ready

src/verifiers/issuerIdentity/didIdentityProof/didIdentityProof.test.ts
```diff
@@ -0,0 +1,33 @@
+import { OpenAttestationDidSignedDidIdentityProof } from "./didIdentityProof";
+
+describe("skip", () => {
+  it("should return skip message", () => {});
+});
+
+describe("test", () => {
```
Nebulis, line 7: I would suggest to follow the structure of other test and to not test separately for test/verify, rather call it through a build in verifier (because that's how verifier are used)

src/verifiers/issuerIdentity/didIdentityProof/didIdentityProof.test.ts
```diff
@@ -0,0 +1,33 @@
...
+  describe("v2", () => {
+    it("should return false for documents not using DID as top level identifier", () => {});
+    it("should return true for documents where any issuer is using the `DID` identity proof", () => {});
+    it("should return true for documents where any issuer is using the `DNS-CONTRACT` identity proof", () => {});
```
Nebulis, line 11: DID-CONTRACT ? (same for v3)

src/verifiers/issuerIdentity/didIdentityProof/didIdentityProof.test.ts
```diff
@@ -0,0 +1,33 @@
+import { OpenAttestationDidSignedDidIdentityProof } from "./didIdentityProof";
```
Nebulis, line 1: it's not stated but it should work the same way for token registry ? and certificate store ?

TODO
```diff
@@ -0,0 +1,14 @@
+Todo
+
+1. Generate sample document for
+    - direct signed by eth (DID)
+    - direct signed by eth and using dns proof (DNS-DID)
```
Nebulis, line 5: DNS-DID or DID-DNS ? I see both used

TODO
```diff
@@ -0,0 +1,14 @@
+Todo
```
Nebulis, line 1: Need to add for revocation

src/verifiers/documentStatus/didSignedDocumentStatus/didSignedDocumentStatus.test.ts
yehjxraymond: ok

src/verifiers/issuerIdentity/didIdentityProof/didIdentityProof.test.ts
yehjxraymond: hmmm i found the structure doing a bit too many things at once, perhaps can have an test around it?

src/verifiers/issuerIdentity/didIdentityProof/didIdentityProof.test.ts
yehjxraymond: right, thanks

TODO
yehjxraymond: the first word is the top level identifier.... so
DNS-DID
DID
DNS-CONTRACT (replacing DNS-TXT)
DID-CONTRACT

TODO
yehjxraymond: right

src/verifiers/issuerIdentity/didIdentityProof/didIdentityProof.test.ts
Nebulis: the problem by separating test / verify is that there is an implicit dependency: when there is a test in the test section returning false, you don't expect the verify section to run the same test because it shouldn't happen.

However if at some point the need change and the test that was previously returning false, returns true, then you would expect a new test to be created in the verify section which can be easily missed.

I don't understand when you say ` i found the structure doing a bit too many things at once`, iirc tests are only running one verifier and checking the fragment ?

PR #115, opened 2020-09-01
fix: handle empty object as input

### LINKED ISSUE #111: Empty object produces an error when passed as document to oa-verify
Hi, noticed that 
```javascript
//  Cannot read property 'issuers' of undefined
const result = await verify({});
```
will throw an error because some validators are unable to find ```issuers``` data inside the empty object.  Prevention of this behavior requires double-checking when validating, and that's not good. I assume that it's better to handle missing keys and just return proper validation error instead of plain JS error.

Nebulis:
closes #111

3 verifiers trigger an error when an empty object is passed down:

OpenAttestationEthereumDocumentStoreStatus
OpenAttestationDnsTxt
OpenAttestationEthereumTokenRegistryStatus
For all those verifiers the error happens during the test function (which determines if a verifier should run or not)

The plan is to fix those test functions so that the verifiers doesn't run if issuers property (in that case) is not provided => the result for those verifiers will be skipped fragment.

On top of that OpenAttestationHash verifier currently return invalid while it shouldn't run because targethash/merkleroot/data are missing => the plan is to fix this behaviour so that the verifier is skipped in that case.

Also to make sure that verifiers test function handle correctly, we should provide a partial object (deeply)


src/verifiers/dnsText/openAttestationDnsTxt.ts
```diff
@@ -44,6 +44,9 @@ const resolveIssuerIdentity = async (
 
 const name = "OpenAttestationDnsTxt";
 const type: VerificationFragmentType = "ISSUER_IDENTITY";
+const isWrappedV2Document = (document: any): document is WrappedDocument<v2.OpenAttestationDocument> => {
```
Nebulis, line 47: write this into open-attestation

src/verifiers/dnsText/openAttestationDnsTxt.ts
```diff
@@ -62,18 +65,20 @@ export const openAttestationDnsTxt: Verifier<
...
   test: (document) => {
-    if (oaUtils.isWrappedV2Document(document)) {
+    if (oaUtils.isWrappedV3Document(document)) {
+      const documentData = getData(document);
+      return documentData.issuer.identityProof.type === v3.IdentityProofType.DNSTxt;
+    } else if (isWrappedV2Document(document)) {
       const documentData = getData(document);
       // at least one issuer uses DNS-TXT
-      return documentData.issuers.some((issuer) => {
+      return documentData?.issuers.some((issuer) => {
```
Nebulis, line 74: add test when missing issuers for all 3 verifiers

gjj (approved):

PR #117, opened 2020-09-02
fix: further tests for empty objects

gjj:


test/fixtures/v2/documentRopstenObfuscatedWithEmptyObject.ts
```diff
@@ -0,0 +1,117 @@
...
+        name: "1894d984-225c-434d-b923-38cb96213ea4:string:GovTech Singapore",
+        url: "09a3cf30-09ea-4ad5-9567-4110d0167620:string:https://tech.gov.sg",
+        documentStore: "6bf70b4b-6af6-41cc-8d7e-bbf47ded03cb:string:0x532C9Ff853CA54370D7492cD84040F9f8099f11B",
+        identityProof: {
+          type: "cb04ba9c-364e-4272-9b90-760842c0e66a:string:DNS-TXT",
+          location: "3376155f-b77f-477b-b50e-483574dee5e9:string:example.openattestation.com",
+        },
+      },
+    ],
+    recipient: {},
```
gjj, line 46: I have two almost identical test certificate here. In this certificate, this is the result after obfuscating the last element (i.e. `recipient.name` etc. is obfuscated individually, rather than the entire `recipient` object).

From https://github.com/Open-Attestation/open-attestation/commit/a0c783ff399f0d8a3390dcf6173c4287a051082d#diff-af044c135a8496a28f0859e897dc301cR108, we expect such an obfuscated certificate to still be valid after `verify`ing it.

But somehow over here, `verify` returns saying it has been tampered. Not sure why though, am I missing something? 😨

gjj: @Nebulis

gjj: Nvm, I realise it was because we haven't upgraded `open-attestation`. I've done so in this MR

Nebulis (approved):

PR #120, opened 2020-09-14
feat: didIdentityProof

yehjxraymond:
Implemented two verification modules:

### didIdentityProof

Checks that all the issuers have signed on the document and return their did as the resolved identity.

### dnsDidProof

Checks that all the issuers with `DNS-DID` has bounded their DNS to the did. 

## Integration Tests

View integration test results: 
https://github.com/Open-Attestation/oa-verify/pull/120/files#diff-89870a61c5e66c74170f57f74e20445dR1158

## TODO (not in first impl)

- Support for other revocation methods other than 'NONE'
- Changing OA / dns-proof package
- DID-CONTRACT support in identity module & document store module


src/verifiers/issuerIdentity/dnsText/openAttestationDnsTxt.ts
```diff
@@ -24,6 +24,7 @@ const resolveIssuerIdentity = async (
   const records = await getDocumentStoreRecords(location);
   const matchingRecord = records.find(
     (record) =>
+      record.addr &&
```
Nebulis, line 27: to fix on dnsprove (addin the comment just for us to keep in mind)

Nebulis: can use optional chain btw

src/verifiers/documentStatus/didSignedDocumentStatus/didSignedDocumentStatus.ts
```diff
@@ -129,10 +55,14 @@ const verify: VerifierType["verify"] = async (_document, _option) => {
...
     if (!document.proof) throw new Error("Document is not signed. Proofs are missing.");
-    const issuanceDeferred: IssuanceStatus[] = issuers.map((issuer: any) =>
+    const signatureVerificationDeferred: DidVerificationStatus[] = issuers.map((issuer: any) =>
       verifySignature({ merkleRoot, identityProof: issuer.identityProof, proof: document.proof, did: issuer.id })
     );
-    const issuance = await Promise.all(issuanceDeferred);
+    const issuance = await (await Promise.all(signatureVerificationDeferred)).map(({ verified, reason, did }) => ({
+      issued: verified,
+      did,
+      ...(reason && { reason }),
```
Nebulis, line 64: https://github.com/Open-Attestation/oa-verify/pull/120/files#diff-43471af6de049beffa58e92b1126431bR43

The reason looks from this ? but this is not an iterator so it will fail

yehjxraymond: This is to conditionally add key. Doing this to not break existing test that doesnt need to change. If I have just the reason key, it will be `reason: undefined`.

image

src/verifiers/issuerIdentity/didIdentityProof/didIdentityProof.ts
```diff
@@ -19,17 +20,63 @@ const skip: VerifierType["skip"] = async () => {
   };
 };
 
-const test: VerifierType["test"] = (_document) => {
-  return true;
+const test: VerifierType["test"] = (document) => {
+  if (!utils.isWrappedV2Document(document)) return false;
+  const { issuers } = getData(document) as any; // TODO Casting to any first to prevent change at the OA level
+  if (issuers.some((issuer: any) => issuer.identityProof?.type === "DID")) return true;
+  return false;
```
Nebulis, line 27:
```suggestion
  return issuers.some((issuer: any) => issuer.identityProof?.type === "DID");
```

src/verifiers/issuerIdentity/didIdentityProof/didIdentityProof.ts
```diff
@@ -19,17 +20,63 @@ const skip: VerifierType["skip"] = async () => {
...
-    status: "VALID",
-  };
+  try {
+    if (!utils.isWrappedV2Document(_document)) throw new Error("Only v2 is supported now");
+    const document = _document as any; // TODO Casting to any first to prevent change at the OA level
+    const data: any = getData(document);
+    const merkleRoot = `0x${document.signature.merkleRoot}`;
+    const signatureVerificationDeferred: Promise<SignatureVerificationFragment>[] = data.issuers.map(
+      async (issuer: any) => {
+        if (issuer.identityProof.type === "DID" || issuer.identityProof.type === "DNS-DID") {
```
Nebulis, line 43: verifier skip DNS-DID

src/verifiers/issuerIdentity/didIdentityProof/didIdentityProof.test.ts
```diff
@@ -1,33 +1,105 @@
...
+              "status": "VALID",
+            },
+          ],
+          "name": "OpenAttestationDidSignedDidIdentityProof",
+          "status": "VALID",
+          "type": "ISSUER_IDENTITY",
+        }
+      `);
+    });
+    it("should fail for documents using `DID` and did signature is not correct", async () => {
```
Nebulis, line 63: I disagree with this test, the signature verification is already made by the document status check, and there is no extra need to do this for the identity (unlike for dns)

test/fixtures/v2/documentDnsDidNoDns.ts
```diff
@@ -0,0 +1,55 @@
+export const documentDnsDidNoDns: any = {
```
Nebulis, line 1: from the test file name I actually expected the location key to not be present, shall we rename documentDnsDidNoDnsTXTEntry or sth similar ?

src/verify.v2.integration.test.ts
```diff
@@ -813,11 +1117,427 @@ describe("verify(integration)", () => {
...
+    expect(results).toMatchInlineSnapshot(`
+      Array [
+        Object {
+          "data": true,
+          "name": "OpenAttestationHash",
+          "status": "VALID",
+          "type": "DOCUMENT_INTEGRITY",
+        },
+        Object {
+          "name": "OpenAttestationSignedProof",
```
Nebulis, line 1171: makes me thing we will remove this one right ?

src/verify.v2.integration.test.ts
```diff
@@ -813,11 +1117,427 @@ describe("verify(integration)", () => {
...
+          "status": "SKIPPED",
+          "type": "ISSUER_IDENTITY",
+        },
+        Object {
+          "data": [TypeError: Cannot read property 'find' of undefined],
+          "name": "OpenAttestationDidSignedDidIdentityProof",
+          "reason": Object {
+            "code": 1,
+            "codeString": "UNEXPECTED_ERROR",
+            "message": "Cannot read property 'find' of undefined",
```
Nebulis, line 1521: :eyes:

src/verifiers/issuerIdentity/didIdentityProof/didIdentityProof.test.ts
yehjxraymond: I disagree with your disagreement. 

There is the need to check for the claim "I am did:xxx:yyy", you perform the check by checking that "did:xxx:yyy" indeed has signed on it. While you perform the same check on the document status, its for different purposes. You will need the module for both `DOCUMENT_STATUS` and `ISSUER_IDENTITY` anyway to pass the final verification.

src/verify.v2.integration.test.ts
yehjxraymond: Yes we should, but i dont want to deal with that in here until this is ready

src/verify.v2.integration.test.ts
yehjxraymond: Well.. it's unexpected.
Guess we'll expect it then

src/verifiers/issuerIdentity/didIdentityProof/didIdentityProof.test.ts
Nebulis: I disagree about your disagreement about my disagreement.

On a more serious note the more we move on, the less I'm convinced that supporting ALL did as identity verifier is not the way to go:
- the exact same work is done twice for 2 different. Just this sounds silly.
- can any DID be used to "verify" identity ? Maybe. Should we support any DID to verify identity ? No
- we have DID-DNS which can provide better identity verification than this. We have been opinionated from the beginning why not continuing and only supporting DNS-DID ?
- we can provide signing with any DID and consumer can provide their own identity verification if they want to
- we can in the future allow some did to perform identity verification (sp/cp ? github?) but I dont think we should allow ALL did

yehjxraymond: Well the problem is that people may not recognize DNS as the best form of identity. 

Using DID, we are telling them, use whatever. If the issuer issues and the recipient accepts, we don't really care. What we want to do is provide the assurance that it has been signed correctly. If you look at the challenges DKIM faces(https://tools.ietf.org/html/rfc6376#section-8), we have them too with DNS-TXT & DNS-DID. 

Like it or not, `example.com`, `Raymond Yeh`, `199201624D`, `S9876543D`, `did:ethr:0xab..cd`, `did:github:nebulis` are all identifiers. Some of them you understand, some you might understand if you google for it, some you might understand and someone else doesnt, and some you can verify. We should be interested in the class of what is verifiable.

yehjxraymond: Ultimately, the user should be educated that they decide whether to accept or not accept documents from a DID, not you. The base layer should be agnostic to this because it is either:
1. Application layer choice (government document verifier might only allow a subset of identifiers)
2. User layer choice (individual choose which identity scheme and identity to believe in)

yehjxraymond: If the base layer could verify it, it should, and the application or user can choose how it wants to deal with it.

yehjxraymond: If you talk about executing the same thing twice, you might want to revisit https://github.com/Open-Attestation/adr/blob/master/verifier.md on the concept of the 3 classes of verification, `DOCUMENT_INTEGRITY`, `DOCUMENT_STATUS` and `ISSUER_IDENTITY`. 

I'm not sure if you disagree at the 3 classes level or at the implementation level.

Nebulis: > Well the problem is that people may not recognize DNS as the best form of identity.
> Using DID, we are telling them, use whatever. If the issuer issues and the recipient accepts, we don't really care. What we want to do is provide the assurance that it has been signed correctly. If you look at the challenges DKIM faces(tools.ietf.org/html/rfc6376#section-8), we have them too with DNS-TXT & DNS-DID.
> Like it or not, example.com, Raymond Yeh, 199201624D, S9876543D, did:ethr:0xab..cd, did:github:nebulis are all identifiers. Some of them you understand, some you might understand if you google for it, some you might understand and someone else doesnt, and some you can verify. We should be interested in the class of what is verifiable.

I agree and they can implement their own identity verification if needed we dont have to cover everything as long as we dont think it's not necessary.
I indeed don't like that `example.com` is thought as identity equally as `did:ethr:0xab..cd` or even `did:github:nebulis` (no I dont like much github as identity too :))
I dont think it's incompatible to allow signing with DID and only recognizing DNS-DID by default.

> Ultimately, the user should be educated that they decide whether to accept or not accept documents from a DID, not you. The base layer should be agnostic to this because it is either:
> Application layer choice (government document verifier might only allow a subset of identifiers)
> User layer choice (individual choose which identity scheme and identity to believe in)
By experience you should know that user will just use whatever works. This strengthen my opinion on we shouldn't allow everything by default. They want to allow specific DID ? Go ahead write your own verifier. They want a verifier to be added into OA ? can discuss

> If you talk about executing the same thing twice, you might want to revisit https://github.com/Open-Attestation/adr/blob/master/verifier.md on the concept of the 3 classes of verification, DOCUMENT_INTEGRITY, DOCUMENT_STATUS and ISSUER_IDENTITY.
> I'm not sure if you disagree at the 3 classes level or at the implementation level.
I'm fine with the 3 classes but DID verification does not so any identity verification imho.

yehjxraymond: So to document our discussion:

L: you used did:xyz ? sorry oa doesn tknow write your own verifier

R: 

the beauty of did is that:
1. You use did:xyz:abc, I dont know what is that but it's giving me a public key that I can check
2. The document has been signed by did:xyz, it can be proven that did:xyz signed on this document
3. Trust this document if:
- a. You trust the xyz verification method (xyz could be corppass)
- b. You trust the abc entity

How is that different from DNS:
1. You use xyz.com, I dont know who is that but it's got a binding to a document store that I can check
2. The document has been issued on the document store. It can be proven that xyz.com signed on this document
3. Trust this document if:
- a. You trust DNS
- b. You trust xyz.com as a company or whatever

L: to me the problem is more to not allow any did by default because it will just cascade everywhere

(by default being the most important word)

R: the whole purpose of did is so to get connected to all 🙈

"interoperability"

L: if it's fully open and we fucked on sth we will be fucked up forever

rjchow: actually I think at the OA level there's only two watertight outcomes:
 - either we completely drop the pretense that we are doing any sort of identity assurance and thus -> we provide a means to retrieve identity metadata but we don't say its an assurance we provide -> issuer_identity verifier should always be skipped unless provided by the consumer (but we can provide a DNS one for them to compose into their solution)
 - we are gatekeepers - this will exclude the acceptance of arbitrary DIDs as issuer identities (its a valid argument to make  that this also excludes arbitrary domain names as issuer identities)

any other combination is leaky and will result in "but then x is not better than y because both of them are the same"

PR #127, opened 2020-10-23
Feat/oa verify issuer identifier impl

yehjxraymond:
This is an incremental change to PR #126 

Things done:
- rename of all identity verifier
- implementation of the 3 identity verifier
- updated test
- exporting new identity verifier builder

Final merge from #126 to master will be marked as breaking change


src/index.ts
```diff
@@ -2,21 +2,18 @@ import { SignedWrappedDocument, v2, v3, WrappedDocument } from "@govtechsg/open-
...
   openAttestationHash,
   openAttestationEthereumTokenRegistryStatus,
   openAttestationEthereumDocumentStoreStatus,
-  openAttestationDnsTxt,
-  OpenAttestationDnsDid,
   OpenAttestationDidSignedDocumentStatus,
+  OpenAttestationIssuerIdentityVerifier,
 ];
 
 const verify = verificationBuilder<
```
rjchow, line 19: This becomes harder to reuse:

```
const x = verifierX; // new identity verifier)

const verify = verificationBuilder([openAttestationHash,
  openAttestationEthereumTokenRegistryStatus,
  openAttestationEthereumDocumentStoreStatus,
  OpenAttestationDidSignedDocumentStatus
  issuerIdentityVerifierBuilder([...defaultIssuerIdentityProofVerifiers, verifierX])])
```

src/verifiers/issuerIdentity/index.ts
```diff
@@ -0,0 +1,27 @@
+import { issuerIdentityVerifierBuilder } from "./builder";
```
rjchow, line 1: Overall,

I find this version much more harder to use than the previous implementation:

- I cant upgrade opencerts to use this version without writing my full own identity implementation (I can't have multiple verifier for one issuer) + validity
- There are different logic for different kind of fragments (I cant mix document_status fragment but I can mix issuer_identity fragment)
- It's not more composable than before (personal opinion but prove me wrong)
- I would not upgrade to this version for opencerts and would rather fix isValid according to my need

src/verifiers/issuerIdentity/index.ts
```diff
@@ -0,0 +1,27 @@
...
+
+// Default set of issuer identity verifiers
+// Currently excludes OpenAttestationDidIdentityProof
+export const defaultIssuerIdentityProofVerifiers = [
+  OpenAttestationDnsDidIdentityProof,
+  OpenAttestationDnsTxtIdentityProof,
+];
+
+// Default ISSUER_IDENTITY verifier, built from defaultIssuerIdentityProofVerifiers
+export const OpenAttestationIssuerIdentityVerifier = issuerIdentityVerifierBuilder(defaultIssuerIdentityProofVerifiers);
```
rjchow, line 16: this bring a difference in how we treat the different kind of verifiers. For instance now we also cant mix different kind of document status (for instance one issuer using document store and one using token registry)

src/verifiers/issuerIdentity/index.ts
```diff
@@ -0,0 +1,27 @@
...
+
+// Exporting all verifiers for dependant apps to compose
+export const verifiers = {
```
rjchow, line 19: Maybe we should be careful with this name because it's exported by the index file and then we have 3 sets of verifiers exported:

- verifiers
- defaultIssuerIdentityProofVerifiers
- openAttestationVerifiers

I fear that verifiers is hard to understand

src/verifiers/issuerIdentity/builder.ts
```diff
@@ -64,6 +39,8 @@ const defaultVerifier: IssuerIdentityVerifier = async ({ issuerIndex }) => {
 export const issuerIdentityVerifierBuilder = (verifiers: IssuerIdentityVerifierDefinition[]): VerifierType => {
   const registeredVerifiers: { [identityProofType: string]: IssuerIdentityVerifier } = {};
   verifiers.forEach((verifier) => {
+    if (registeredVerifiers[verifier.type])
+      throw new Error(`Cannot register verifier to a type ${verifier.type} which has existing verifier assigned`);
     registeredVerifiers[verifier.type] = verifier.verify;
   });
```
rjchow, line 46: if .test() is always true then skipFragment should never be called if skipFragment is called it implies something has gone wrong so perhaps it should throw an error to indicate that the world has ended in an unexpected way

rjchow: Using the verification based on the type makes it completely invalid / unusable for OpenCerts:

- the registry verifier run for all documents
- DNS-TXT may run as well for this documents

That means one issuer may have 2 verifications but this implementation doesn't allow for it. Registry verifier has priority over DNS-TXT. A new institution will need to use DNS-TXT as long as they are not part of the registry so this will be needed forever.

On top of that there is the case where one verifier may fail.

src/verifiers/issuerIdentity/builder.ts
```diff
@@ -89,7 +66,12 @@ export const issuerIdentityVerifierBuilder = (verifiers: IssuerIdentityVerifierD
         const documentData = getData(document);
         const verificationResultsDeferred = documentData.issuers.map((issuer, issuerIndex) => {
           const identityProofType = issuer.identityProof?.type;
-          if (!identityProofType) throw new Error("");
+          if (!identityProofType)
```
rjchow, line 69: We dont check for this scenario (identityProofType) in v3.

src/index.ts
yehjxraymond: Any suggestions?
I think we need to separate ease of reuse and correctness. Also, I think a sensible default will help with the ease of reuse.

src/verifiers/issuerIdentity/index.ts
yehjxraymond: The problem with this actually go away in V3, the problem was that our method for verifying multiple identities in V2 is incorrect (not inelegant). 

When you run multiple verifiers against a list of identities it is impossible to ensure the validity of the issuers if they use different schemes. 

Ie. 
Issuer 1 - DNS-TXT
Issuer 2 - Opencerts Registry
Issuer 3 - Non-existent method

Having both Opencerts registry and Dns-txt will skip past the last issuer and return VALID where it should have been INVALID when *not all* the issuers are verified. 

This problem goes away in V3 when there is only one issuer and that we can run all the independent fragments like before... I'll describe the upgrade process for OpenCerts registry in the comment below.

src/verifiers/issuerIdentity/builder.ts
yehjxraymond: On OpenCerts, the method will be to:

1. Not use the DNS-TXT module. Instead create a module that inherits from the DNS-TXT and do an OR with the OpenCerts Registry:

```
  const dnsTxtResults = verify({document, issuerIndex, options});
  const registryResults = getRegistryIdentity(documentStore);
  return {
    status: valid if any of result is valid,
    identifier: registryResult takes precedence, else dnsTxtResult's identity
  }
```

2. Export this new identity provider

```
export const OpenCertsDnsTxtIdentityProof: IssuerIdentityVerifierDefinition = {
  type: "OC-DNS-TXT",
  verify,
};
```

3. Use it:

```
  const identityVerifier = issuerIdentityVerifierBuilder([OpenCertsDnsTxtIdentityProof, OpenAttestationDnsDidIdentityProof]);
  const verify = verificationBuilder([
  openAttestationHash,
  openAttestationEthereumTokenRegistryStatus,
  openAttestationEthereumDocumentStoreStatus,
  OpenAttestationDidSignedDocumentStatus,
  identityVerifier
])
```

src/verifiers/issuerIdentity/builder.ts
yehjxraymond: v3 does not allow for undef types, v2 has that for legacy reasons (OpenCerts)

src/verifiers/issuerIdentity/index.ts
yehjxraymond: Conversation Continued:

Rui (Govtech), [27.10.20 15:29]
probably better to do the discussion on the github thread because if not need to copypaste to laurent again since the comments came from him

Rui (Govtech), [27.10.20 15:29]
there's an issue with correctness

Rui (Govtech), [27.10.20 15:30]
because way back in the day we made that original sin of deciding that oneOf(identity) is sufficient to be valid

Rui (Govtech), [27.10.20 15:30]
actually that's not correct either

Raymond Yeh, [27.10.20 15:30]
[In reply to Rui (Govtech)]
ehh we didnt decide on this

Rui (Govtech), [27.10.20 15:30]
its more like

if (registry || oneOf(other identities))

Raymond Yeh, [27.10.20 15:30]
it was suppose to be allOf but i think because we didnt think there were more than one way to verify identity

Rui (Govtech), [27.10.20 15:31]
nah we made that decision i remember we had that conversation

Raymond Yeh, [27.10.20 15:31]
if (registry || oneOf(other identities))
^ for single issuer you see....

Raymond Yeh, [27.10.20 15:31]
so should be 
allOf ( (registry || oneOf(other identities)) )

Raymond Yeh, [27.10.20 15:32]
we had not been thinking about the case about (multiple issuer) x (different way to verify)

Rui (Govtech), [27.10.20 15:32]
:D.........

Raymond Yeh, [27.10.20 15:33]
[In reply to Rui (Govtech)]
also, this is reduced to 
(registry || verifier 1 || verifier 2 || etc) 
^ which is the premise of the identity builder

yehjxraymond:
Closing in favor of stricter checks on individual verifier to not allow mixing of issuers

PR #129, opened 2020-10-30
feat: restrict token registry to fail on multiple issuers

yehjxraymond:
Having stricter check with token registry to disallow matching it with other identity verification types. 

- verifier to fail if its not ONE issuer that is using `tokenRegistry`.
- refactored catch-all error handler to be used across all verifiers. simply use `withCodedErrorHandler` to wrap verify function and throw know errors with `CodedError`.


src/common/errorHandler.ts
```diff
@@ -0,0 +1,46 @@
...
+  unexpectedErrorCode: number;
+  unexpectedErrorString: string;
+}
+
+export const withCodedErrorHandler = (verify: Verifier["verify"], errorOptions: ErrorOptions) => async (
+  document: Parameters<Verifier["verify"]>[0],
+  options: Parameters<Verifier["verify"]>[1]
+) => {
+  try {
+    // Using return await to ensure async function execute in try loop
```
Nebulis, line 15:
```suggestion
    // Using return await to ensure async function execute in try block
```

src/verifiers/documentStatus/tokenRegistryStatus/openAttestationEthereumTokenRegistryStatus.ts
```diff
@@ -1,22 +1,91 @@
...
+  tokenRegistry,
+  merkleRoot,
+  network,
+}: {
+  tokenRegistry: string;
+  merkleRoot: string;
+  network: string;
+}): Promise<{ minted: boolean; reason?: string }> => {
+  try {
+    const tokenRegistryContract = await TradeTrustErc721Factory.connect(tokenRegistry, getProvider({ network }));
```
Nebulis, line 63: - why getting rid of code and codeString ? that makes it a breaking change ...
- why getting rid of SERVER_ERROR ? We use it on opencert to differentiate a business and a technical error

yehjxraymond: I knew you will ask, which is why i got you to review. 
For the errors which this verifier *knows for sure* that the result is either true or false, it returns that. 
For errors which it is uncertain if the document is issued or not, it will throw. 
There is no meaning in returning more information than that, oa-verify is meant to provide that level of abstraction to dependant applications.
In the cases from line 69 to 82, we are certain the token was not minted (regardless of which error it is), there is **no error** in those cases. 

I was initially reluctant to return more information that true/false.

yehjxraymond: As for `SERVER_ERROR`, I can make the change to throw that one with a code.

PR #139, opened 2020-11-02
feat: aligning naming convention with files and modules

yehjxraymond:


src/index.ts
```diff
@@ -2,21 +2,21 @@ import { SignedWrappedDocument, v2, v3, WrappedDocument } from "@govtechsg/open-
...
-import { OpenAttestationDidSignedDidIdentityProof } from "./verifiers/issuerIdentity/didIdentityProof";
-import { OpenAttestationDnsDid } from "./verifiers/issuerIdentity/dnsDidProof";
+import { openAttestationEthereumTokenRegistryStatus } from "./verifiers/documentStatus/tokenRegistry";
+import { openAttestationEthereumDocumentStoreStatus } from "./verifiers/documentStatus/documentStore";
+import { openAttestationDidSignedDocumentStatus } from "./verifiers/documentStatus/didSigned";
+import { Identity, openAttestationDnsTxtIdentityProof } from "./verifiers/issuerIdentity/dnsTxt";
+import { openAttestationDidIdentityProof } from "./verifiers/issuerIdentity/did";
+import { openAttestationDnsDidIdentityProof } from "./verifiers/issuerIdentity/dnsDid";
 
 const openAttestationVerifiers: Verifiers[] = [
```
Nebulis, line 13: what's the convention followed for the verifiers name ?

yehjxraymond: All the identityProof ones are not postfixed with `IdentityProof` while document status will be postfixed by `Status`

yehjxraymond: before that we have DnsText, but that was the only one, so we added more now, all of them are postfixed with IdentityProof to distinguish

PR #141, opened 2020-11-18
feat: verificationBuilder with custom provider

yehjxraymond:


src/verifiers/documentStatus/tokenRegistry/ethereumTokenRegistryStatus.ts
```diff
@@ -57,15 +56,24 @@ export const getTokenRegistry = (
...
+    error.method?.toLowerCase() === "ownerOf(uint256)".toLowerCase() &&
+    error.code === errors.CALL_EXCEPTION
+  );
+};
 const decodeError = (error: any) => {
   const reason = error.reason && Array.isArray(error.reason) ? error.reason[0] : error.reason ?? "";
   switch (true) {
-    case reason.toLowerCase() === "ERC721: owner query for nonexistent token".toLowerCase() &&
-      error.code === errors.CALL_EXCEPTION:
+    case isNonExistentToken(error):
```
Nebulis, line 74: Hummmm that means that we will need to upgrade to this new version for this specific fix

I guess it's ok it's a very specific case, at worst it will throw a different error

src/verifiers/verificationBuilder.ts
```diff
@@ -1,21 +1,31 @@
...
  * Before running each verifier, the manager will make sure the verifier can handle the specific document by calling its exposed test function.
  * The manager will return the consolidated list of {@link VerificationFragment}
  */
-export const verificationBuilder = <Document>(verifiers: Verifier<Document>[]) => (
-  document: Document,
-  options: VerificationManagerOptions
-): Promise<VerificationFragment[]> => {
+export const verificationBuilder = <Document>(
+  verifiers: Verifier<Document>[],
+  builderOptions: VerificationBuilderOptions
```
Nebulis, line 17: Isn't this a breaking change?

Nebulis, line 17: This is a breaking change.

Now that I think about it even the change of signature of verifier is a breaking change (ie for the one using a custom verifier this won't work anymore)

README.md
```diff
@@ -82,9 +82,27 @@ console.log(isValid(fragments)); // display true
 
 ### Environment Variables
 
-- `ETHEREUM_PROVIDER`: let you pick the provider you want to use. Available values: `cloudflare`. The provider will default to `infura` if the variable is not set.
```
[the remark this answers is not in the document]
yehjxraymond: Yes I intend to add a breaking change flag to this PR, cause it changes the fn signature of the builder and verify already

src/common/utils.ts
```diff
@@ -1,7 +1,11 @@
-import * as ethers from "ethers";
+import { providers } from "ethers";
+import { VerificationBuilderOptions, VerificationBuilderOptionsWithNetwork } from "../types/core";
 import { INFURA_API_KEY } from "../config";
 
-export const getProvider = (options: { network: string }): ethers.providers.Provider =>
-  process.env.ETHEREUM_PROVIDER === "cloudflare"
-    ? new ethers.providers.CloudflareProvider()
-    : new ethers.providers.InfuraProvider(options.network, INFURA_API_KEY);
+export const getDefaultProvider = (options: VerificationBuilderOptionsWithNetwork): providers.Provider => {
```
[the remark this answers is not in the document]
yehjxraymond: Yes, use a fallback provider, there is an implementation example in the ether's getDefaultProvider, it's creating a mix of provider already. I will actually suggest that we implement with one infura and one alchemy key. I've heard good things about alchemy. 

The only problem is that... alchemy's error might look different from infura, will have to test it with the error decoders for docuemntstore and tokenregistry.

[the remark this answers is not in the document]
yehjxraymond: I dont understand your second question though.. any examples?

src/common/utils.ts
```diff
@@ -1,7 +1,11 @@
...
+  return new providers.InfuraProvider(options.network, INFURA_API_KEY);
+};
+
+export const getProvider = (options: VerificationBuilderOptions): providers.Provider => {
+  return "provider" in options ? options.provider : getDefaultProvider(options);
```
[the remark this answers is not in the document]
yehjxraymond: Haha cannot, it doesnt have the type for `provider` before restricting the interface to the one with provider... stupid i know.

src/verifiers/documentStatus/tokenRegistry/ethereumTokenRegistryStatus.ts
yehjxraymond: I honestly dont know a better way to go about this. I hate that there is so much coupling with the infura's error message...

src/verifiers/verificationBuilder.ts
yehjxraymond: Yes

src/common/utils.ts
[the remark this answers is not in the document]
Nebulis: ah ok .. :)

src/verifiers/verificationBuilder.ts
yehjxraymond: Yes.. thats why all the test files are updated. The design considerations behind this change is that when we define or build the verify function we know which network it is on already, so it shouldnt be the onus of the `verify` caller to define the network again, this way you wont have the `network` strewn all over the caller code.

src/verifiers/documentStatus/tokenRegistry/ethereumTokenRegistryStatus.ts
Nebulis: It's ok for this MR, I mainly mean for our different tools, it's not a big deal I htink, but sth to keep in mind

Nebulis (approved):

PR #152, opened 2021-01-25
feat: oav3 document store

yehjxraymond:


scripts/generate.v3.ts
```diff
@@ -24,10 +27,11 @@ const ethereumDocumentConfig = {
     address: "0x1245e5B64D785b25057f7438F715f4aA5D965733",
     key: "0x416f14debf10172f04bef09f9b774480561ee3f05ee1a6f75df3c71ec0c60666",
   },
-  timeout: 60 * 1000, // 1 min timeout for contract to execute
+  timeout: 5 * 60 * 1000, // 5 min timeout for contract to execute
```
Nebulis, line 30: 1min not enough ? xD

scripts/generate.v3.ts
```diff
@@ -2,13 +2,16 @@
 import { __unsafe__use__it__at__your__own__risks__wrapDocument, v3 } from "@govtechsg/open-attestation";
```
Nebulis, line 2: do we need to commit the generated files ?

src/verifiers/documentStatus/documentStore/ethereumDocumentStoreStatus.ts
```diff
@@ -217,97 +218,147 @@ export const isRevokedOnDocumentStore = async ({
...
 };
-export const openAttestationEthereumDocumentStoreStatus: Verifier<
-  WrappedDocument<v2.OpenAttestationDocument> | WrappedDocument<v3.OpenAttestationDocument>
-> = {
-  skip: () => {
-    return Promise.resolve({
-      status: "SKIPPED",
-      type,
+
+const test: VerifierType["test"] = (document) => {
```
Nebulis, line 235: just to add that isWrappedV3Document is not correct as well (it doesn't do what you expect). We might need to discuss if we actually want to have such a helper in our lib :)

scripts/generate.v3.ts
yehjxraymond: yes pls, it takes a while to generate all of them, its just to generate test fixtures infrequently...

scripts/generate.v3.ts
yehjxraymond: Its enough for now, but have seen cases where the network choked

src/verifiers/documentStatus/documentStore/ethereumDocumentStoreStatus.ts
yehjxraymond: "yes"

Nebulis (approved):

PR #157, opened 2021-02-02
fix: token registry's test & oav3 integration test

yehjxraymond:


src/verify.v3.integration.test.ts
```diff
@@ -1,7 +1,574 @@
...
+            "name": "OpenAttestationDidIdentityProof",
+            "status": "VALID",
+            "type": "ISSUER_IDENTITY",
+          },
+        ]
+      `);
+      expect(valid).toBe(true);
+    });
+  });
+  describe("invalid documents", () => {
```
Nebulis, line 444: for this failing test, maybe we could check the snapshot of the fragments that we expect to fail, only:
```
const fragment = fragments.find(ble => ble)
expect(fragment).toMatchInlineSnapshot()
```

something like this. (I think you decided to avoid to had to many snapshots)

yehjxraymond: good idea, i'll add one: `findFailingFragment` later

Nebulis (approved):

PR #161, opened 2021-02-25
feat: add getIdentifier and getIdentityFragment util

cavacado:
context: https://www.pivotaltracker.com/story/show/176797200
this pr does the following:
- adds `getIdentifier` and `getIdentityFragment` utils
- adds istanbul coverage folder to .gitignore

### DIFF
--- a/src/getIdentifier.test.ts
+++ b/src/getIdentifier.test.ts
@@ -84,23 +84,20 @@ const malformedVerificationFragment2: VerificationFragment = {
   status: "VALID",
 };
 
-
 describe("getIdentifier", () => {
   it("should throw an error when no fragments are provided", () => {
     expect(() => getIdentifier([])).toThrowError("Please provide at least one verification fragment");
   });
   it("should return an unknown identity proof, when fragment.name is out of specification", () => {
-    expect(
-      getIdentifier([verificationFragment1, malformedVerificationFragment1])
-    ).toStrictEqual({
+    expect(getIdentifier([verificationFragment1, malformedVerificationFragment1])).toStrictEqual({
       identifier: "Unknown",
       type: "Unknown",
     });
   });
   it("should throw an error when fragment.data cannot be handled", () => {
-    expect(() =>
-      getIdentifier([verificationFragment1, malformedVerificationFragment2]))
-    .toThrowError("No data property found in fragment, malformed fragment");
+    expect(() => getIdentifier([verificationFragment1, malformedVerificationFragment2])).toThrowError(
+      "No data property found in fragment, malformed fragment"
+    );
   });
 
   describe("v2", () => {
--- a/src/getIdentifier.test.ts
+++ b/src/getIdentifier.test.ts
@@ -1,5 +1,5 @@
 import { v3 } from "@govtechsg/open-attestation";
-import { getIdentityProofFragment, getIdentifier } from "./getIdentifier";
+import { getIdentifier } from "./getIdentifier";
 import { VerificationFragment } from "./types/core";
 import {
   openAttestationDnsTxtIdentityProof,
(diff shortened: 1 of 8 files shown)


cavacado:
btw do we have a `compose` util anywhere in our repos?
also could the reviewer/s please scrutinise the error states that could happen?
`fragment.data` is untyped so I might miss out some edge case

src/getIdentifier.ts
```diff
@@ -0,0 +1,81 @@
...
+      type,
+    }));
+  }
+  return {
+    identifier: fragment.data.did,
+    type,
+  };
+};
+
+export const getIdentityProofFragment = (fragments: VerificationFragment[]) => {
```
Nebulis, line 54: we already have some types for some fragment, maybe we can enhance and add runtypes as well, so that we are sure about what we handle

src/getIdentifier.ts
```diff
@@ -0,0 +1,81 @@
...
+  if (fragments.length < 1) {
+    throw new Error("Please provide at least one verification fragment");
+  }
+  return fragments.find((status) => status.type === "ISSUER_IDENTITY" && status.status === "VALID");
+};
+
+export const getIdentifier = (fragment: ReturnType<typeof getIdentityProofFragment>) => {
```
Nebulis, line 61: I didn't clone locally, but is the return type correct ? For instance we could expect type tp be either "DNS" | "DID" | "DNS-DID" | "unknown". I think it will be inferred as string (I might be wrong too)

src/getIdentifier.ts
```diff
@@ -0,0 +1,81 @@
...
+  if (typeof fragment === "undefined") {
```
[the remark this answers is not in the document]
cavacado: makes sense, will change

src/getIdentifier.ts
cavacado: have added more specific types for each of the functions so that we are sure what we are handling

src/getIdentifier.ts
cavacado: return type is `VerificationFragment<any> | undefined`, which will be further refined after the guards so to me its robust enough.

src/getIdentifier.test.ts
```diff
@@ -0,0 +1,151 @@
...
+
+describe("getIdentityProofFragment", () => {
+  it("should throw an error when no fragments are provided", () => {
+    expect(() => getIdentityProofFragment([])).toThrowError("Please provide at least one verification fragment");
+  });
+});
+
+describe("getIdentifier", () => {
+  it("should throw an error when undefined is passed as a fragment", () => {
+    expect(() => getIdentifier(getIdentityProofFragment([verificationFragment1]))).toThrowError(
```
Nebulis, line 89: it's not abvious where undefined is, let's just provide undefined directly

cavacado: 👌

src/getIdentifier.test.ts
```diff
@@ -0,0 +1,151 @@
...
+  type: "ISSUER_IDENTITY",
+  data: [
+    {
+      did: "did:ethr:0xE712878f6E8d5d4F9e87E10DA604F9cB564C9a89",
+      status: "VALID",
+    },
+  ],
+  status: "VALID",
+};
+const malformedVerificationFragment1: VerificationFragment = {
```
Nebulis, line 68: let's add a comment to explain why this fragment is malformed, same for the other

src/getIdentifier.ts
```diff
@@ -0,0 +1,92 @@
...
+      type,
+    }));
+  }
+  return {
+    identifier: fragment.data?.did,
+    type,
+  };
+};
+
+export const getIdentityProofFragment = (fragments: VerificationFragment[]) => {
```
Nebulis, line 65: why do we need this function ? why isn't this function called directly within getIdentifier ?

cavacado: no particular reason, wanted to segregate the errors caused by wrong data passed in; instead of handling everything in getIdentifier

src/getIdentifier.test.ts
cavacado: sure

Nebulis (approved):

Issue #162, opened 2021-02-25
Verifying did identity with V3 does not seem to support custom RPC provider


jaffaskree:
I am trying to verify documents using V3 and a custom RPC provider (having network id "cyan-imda").

1. _verificationBuilder_ takes an array of verifiers and some options. Thanks to these options, we can pass a custom ethers provider.

2. For each verifier, it calls verifier.verify(document, verifierOptions), so we can assume that our custom provider is passed to each verifier.

3. But looking at didIdentityProof verifier, we see that it calls verifyV3 without passing our custom provider.

4. Then verifyV3 calls verifySignature().

5. verifySignature immediately calls getPublicKey().

6. getPublicKey immediately calls resolve().

7. resolve tries to use the default resolver, which is configured with the default network and default RPC provider.

==> the error is thrown: "No conf for networkId: cyan-imda".
==> the error is propagated and finaly catched by the error wrapper withCodedErrorHandler.
==> verification fails.

Nebulis:
Hi, thanks for the detailed report.

I guess you would expect a resolver to be built [that way](https://github.com/decentralized-identity/ethr-did-resolver#multi-network-configuration):
```
const providerConfig = {
  networks: [
    { name: "whatever", provider: ethersProvider }
  ]
}
```

We never tried to use ethers provider with `ethr-did-resolver` and the example mention only about web3. So I'm not sure whether ethers is fully supported but we can try.

However, provider is mainly meant for document store and token registry verification. For DID I think we should introduce a new resolver option maybe.

hitesh-sm:
Please correct me if I am wrong Benoit @jaffaskree , I think the main problem we are facing here is that the provider is hardcoded to mainnet 
https://github.com/Open-Attestation/oa-verify/blob/ad06d88e963e95464c349af4f2c2dbb1bc07cb6d/src/did/resolver.ts#L8

This would mean that calls for resolving DID for any other network will fail. Can we instead pass the provider from options and only default to mainnet when there is no provider present in the options.

jaffaskree:
@hitesh-sm Yes.

@Nebulis I didn't mean that the provider should necessary be ethers-like. Using ethers provider is not really the problem. 
What I had in mind is: it would be great to be able to pass our own config to the resolver in order to use custom network.

Currently, when calling verificationBuilder, the custom config seems to be dropped at the time you wrap the verify function inside withCodedErrorHandler.

[...]

PR #163, opened 2021-02-26
feat: provide custom resolver option for did

jaffaskree:
In response to #162. Basically, it adds the possibility to pass an optional custom did resolver to the verificationBuilder options, allowing to resolve dids on custom chains.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2272,6 +2272,7 @@
         "jest-resolve": "^26.1.0",
         "jest-util": "^26.1.0",
         "jest-worker": "^26.1.0",
+        "node-notifier": "^7.0.0",
         "slash": "^3.0.0",
         "source-map": "^0.6.0",
         "string-length": "^4.0.1",
@@ -6363,6 +6364,7 @@
       "integrity": "sha512-yAYxeGpVi27hqIilG1nh4A9Bnx4J3Ov+eXy4koL3drrR+IO9GaWPsKjik20ht608Asqi8TQPf0mczhEeyAtMzg==",
       "dev": true,
       "dependencies": {
+        "@commitlint/load": ">6.1.1",
         "chalk": "^2.4.1",
         "commitizen": "^4.0.3",
         "conventional-commit-types": "^3.0.0",
@@ -7179,7 +7181,8 @@
         "esprima": "^4.0.1",
         "estraverse": "^4.2.0",
         "esutils": "^2.0.2",
-        "optionator": "^0.8.1"
+        "optionator": "^0.8.1",
+        "source-map": "~0.6.1"
       },
       "bin": {
         "escodegen": "bin/escodegen.js",
@@ -9281,6 +9284,7 @@
         "minimist": "^1.2.5",
         "neo-async": "^2.6.0",
         "source-map": "^0.6.1",
+        "uglify-js": "^3.1.4",
         "wordwrap": "^1.0.0"
       },
       "bin": {
@@ -11278,6 +11282,7 @@
         "@types/graceful-fs": "^4.1.2",
         "anymatch": "^3.0.3",
         "fb-watchman": "^2.0.0",
+        "fsevents": "^2.1.2",
         "graceful-fs": "^4.2.4",
         "jest-serializer": "^26.1.0",
         "jest-util": "^26.1.0",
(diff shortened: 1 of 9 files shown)


package-lock.json
```diff
@@ -2272,7 +2272,6 @@
         "jest-resolve": "^26.1.0",
```
Nebulis, line 2272: please revert this file

src/did/resolver.ts
```diff
@@ -1,29 +1,38 @@
...
 
 const didResolutionCache = new NodeCache({ stdTTL: 5 * 60 }); // 5 min
 
-export const resolver = new Resolver({
-  ...ethrGetResolver(providerConfig),
-  ...webGetResolver(),
-});
+const defaultResolver = new Resolver({ ...ethrGetResolver(providerConfig), ...webGetResolver() });
+
+export const createResolver = (ethrResolverConfig: EthrResolverConfig): Resolver => {
```
Nebulis, line 23: the naming bother me a bit, because it's a generic name to create a custom ethr did. Suggestion
- rename to createEthrDidResolver
- make the first param an object, with `ethrResolverConfig` as optional and then add or not `ethrGetResolver(ethrResolverConfig)` depending on the presence of the option.

src/did/resolver.ts
```diff
@@ -1,29 +1,38 @@
-import { Resolver, DIDDocument } from "did-resolver";
+import { Resolver, DIDDocument, PublicKey } from "did-resolver";
```
Nebulis, line 1: let's add a test to make sure the custom resolver works. One test will be enough, we can use msw + fake url to mock the response

src/did/resolver.ts
jaffaskree: 1) The name createEthrDidResolver could be confusing, as it creates a resolver both for ethr method and web method, like the default resolver.

2) If the custom ethr config is optional, I guess that if it's not passed as args, I should return the default resolver. So, what do you think of:

```
export const createResolver = (ethrResolverConfig?: EthrResolverConfig): Resolver => {
  return ethrResolverConfig
    ? new Resolver({ ...ethrGetResolver(ethrResolverConfig), ...webGetResolver() })
    : defaultResolver;
};
```

src/did/resolver.ts
jaffaskree: I tried to implement a mock result for a custom resolver, but it was very complicated because the resolver makes many JSON-RPC calls that are not easily mockable with a REST-dedicated tool like msw. 

After a few hours, I gave up. Not only because it's overcomplicated, but also because I realized that it didn't make sense. It's like testing the ethr-did-resolver library itself.

At this point, my conclusion is:

1. You have already tested all the verifiers with a resolver using Infura provider.
2. The goal of this PR is to allow passing another resolver.
3. Testing a custom resolver doesn't make sense, because "resolve" operations rely on external libs (did-resolver and ethr-did-resolver) and those libs are already tested.
4. Or, maybe I didn't understand what you expected me to test ;)

package-lock.json
jaffaskree: done

src/did/resolver.ts
Nebulis: I get what you mean, I will try on my side tomorrow and if I came to the same conclusion as your we will proceed as is

src/did/resolver.ts
Nebulis: Yes that will do, but also I think it would be better to use an object, so that it will be easier to add more resolver configuration later

```ts
export const createResolver = ({ethrResolverConfig}: {ethrResolverConfig?: EthrResolverConfig}): Resolver => {
  return ethrResolverConfig
    ? new Resolver({ ...ethrGetResolver(ethrResolverConfig), ...webGetResolver() })
    : defaultResolver;
};
```

src/did/resolver.ts
Nebulis: Ok so I confirm there are internally at least 3 different calls to the URL provided. Let's not use msw.

We can however resolve a ropsten specific DID. The following DID `did:ethr:ropsten:0xf3beac30c498d9e26865f34fcaa57dbb935b0d74` will not resolve with the default resolver. We can create a test that provides a ropsten resolver. Here is an example by calling the resolver function directly:

```ts
    const resolver = createResolver({
      networks: [{ name: "ropsten", rpcUrl: `https://ropsten.infura.io/v3/${INFURA_API_KEY}` }],
    });
    const did = await resolver.resolve("did:ethr:ropsten:0xf3beac30c498d9e26865f34fcaa57dbb935b0d74");
```

Let's do it directly on the did verifier

jaffaskree: Ok, I successfully tested it using directly the resolve() function. But if I want to use the verifier instead (openAttestationDidIdentityProof), I need a signed document to verify. For that, I need the private key of did:ethr:ropsten:0xf3beac30c498d9e26865f34fcaa57dbb935b0d74, in order to sign the doc.

jaffaskree: Ok, so I created a new ropsten address to have a fresh new keypair. Then I tested as you proposed. It's ok, but there is a remaining problem : I can test successful scenario (with the custom resolver), but not failure scenario (without the custom resolver). Why ? Because when a did successfully resolved, it is added to a cache for 5 minutes. So, I cannot "expect" a failure when the custom resolver is not passed because the did document is retrieved from the cache. Of course, if I disable the cache by commenting that part of the code, it works. I don't think I should modify the resolver again to allow disabling the cache, so I'm going to push my tests with successful scenario only.

jaffaskree: https://github.com/Open-Attestation/oa-verify/pull/163/commits/03f487d8620b8f6b198405fb876001fb18e5cbf6

Nebulis: I'm fine with the working path only for the moment. Thanks :)

src/did/resolver.ts
jaffaskree: done

Nebulis (approved):

Nebulis:
hummm it doesnt run the ci :cold_sweat:

Nebulis:
will merge and fix from master if needed

PR #166, opened 2021-03-09
fix: util get obfuscated

simboonlong:
- https://www.pivotaltracker.com/story/show/176797231

### DIFF
--- a/src/common/utils.test.ts
+++ b/src/common/utils.test.ts
@@ -1,5 +1,5 @@
 import { v2, SchemaId, WrappedDocument } from "@govtechsg/open-attestation";
-import { getObfuscated } from "./utils";
+import { isObfuscated, getObfuscatedData } from "./utils";
 
 const documentRopstenNotObfuscated: WrappedDocument<v2.OpenAttestationDocument> = {
   version: SchemaId.v2,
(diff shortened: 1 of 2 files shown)


src/common/utils.test.ts
```diff
@@ -64,15 +64,20 @@ const documentRopstenObfuscated: WrappedDocument<v2.OpenAttestationDocument> = {
...
   it("should return true where there is obfuscated data in v2 document", () => {
-    const { isOfuscated, obfuscated } = getObfuscated(documentRopstenObfuscated);
-    expect(isOfuscated).toBe(true);
-    expect(obfuscated?.length).toBe(1);
-    expect(obfuscated?.[0]).toBe("45c49a0e6efbde83c602cf6bbe4aa356d495feaf78a9a309cc1bad101f5c52f4");
+    expect(isObfuscated(documentRopstenObfuscated)).toBe(true);
+  });
+
+  it("should return undefined when there is no obfuscated data in v2 document", () => {
+    expect(getObfuscatedData(documentRopstenNotObfuscated)).toBe(undefined);
```
Nebulis, line 75: should we return an empty array instead ? I'm not fond of the undefined. wdyt ?

src/common/utils.ts
```diff
@@ -11,27 +11,34 @@ export const getProvider = (options: VerificationBuilderOptions): providers.Prov
...
+  if (utils.isWrappedV3Document(document)) {
+    return !!document.proof.privacy?.obfuscated?.length;
+  }
+
+  if (utils.isWrappedV2Document(document)) {
+    return !!document.privacy?.obfuscatedData?.length;
+  }
+
+  throw new Error(
+    "Unsupported document type: Only can retrieve isObfuscated from wrapped OpenAttestation v2 & v3 documents."
```
Nebulis, line 26: suggestion: Can only check if there are obfuscated data from wrapped OpenAttestation v2 & v3 documents

src/common/utils.ts
```diff
@@ -11,27 +11,34 @@ export const getProvider = (options: VerificationBuilderOptions): providers.Prov
...
-    return {
-      isOfuscated: !!document.privacy?.obfuscatedData?.length,
-      obfuscated: document.privacy?.obfuscatedData,
-    };
+    return document.privacy?.obfuscatedData;
   }
 
-  return { isOfuscated: false, obfuscated: undefined };
+  throw new Error(
+    "Unsupported document type: Only can retrieve obfuscated data from wrapped OpenAttestation v2 & v3 documents."
```
Nebulis, line 42: Can only ?

src/common/utils.test.ts
simboonlong: no opinion on this, yeah i can return empty array

Nebulis (approved):

PR #167, opened 2021-03-19
chore: improve fragment types

Nebulis:
### Overview

Various utilities and types are available to assert the correctness of fragments. Each verification method exports types for the fragment, and the data associated with the fragment.

- fragment types are available in 4 flavors: `VALID`, `INVALID`, `SKIPPED`, and `ERROR`.
- `VALID` and `INVALID` fragment data are available in 2 flavors most of the time, one for each version of `OpenAttestation`.

This library provides types and utilities to:

- get a specific fragment from all the fragments returned by the `verify` method
- narrow down to a specific type of fragment
- narrow down to a specific fragment data

Let's see how to use it

### Example

```ts
import {utils} from "@govtechsg/oa-verify";
const fragments = verify(documentRopstenValidWithCertificateStore, { network: "ropsten" });
// return the correct fragment, correctly typed
const fragment = utils.getOpenAttestationEthereumTokenRegistryStatusFragment(fragments)
if(utils.isValidFragment(fragment)) {
    // guard to narrow to the valid fragment type
    const {data} = fragment;
    if (ValidTokenRegistryDataV2.guard(data)) {
        // data is correctly typed here
    }
}
```

Note that in the example above, using `utils.isValidFragment` might be unnecessary. It's possible to use directly `ValidTokenRegistryDataV2.guard` over the data.

### List of utilities
- `getOpenAttestationHashFragment`
- `getOpenAttestationDidSignedDocumentStatusFragment`
- `getOpenAttestationEthereumDocumentStoreStatusFragment`
- `getOpenAttestationEthereumTokenRegistryStatusFragment`
- `getOpenAttestationDidIdentityProofFragment`
- `getOpenAttestationDnsDidIdentityProofFragment`
- `getOpenAttestationDnsTxtIdentityProofFragment`
- `isValidFragment`: type guard to filter only `VALID` fragment type
- `isInvalidFragment`: type guard to filter only `INVALID` fragment type
- `isErrorFragment`: type guard to filter only `ERROR` fragment type
- `isSkippedFragment`: type guard to filter only `SKIPPED` fragment type

===================================================================================================
I'm not stuck anymore. I created (more or less) correct types for the hash and dns verifiers. I made changes on the type inference for the verification builder.

You can clone locally and play with types in the following files (https://github.com/Open-Attestation/oa-verify/pull/167/files#diff-a2a171449d862fe29692ce031981047d7ab755ae7f84c707aef80701b3ea0c80R41). The best I can suggest so far for this implementation is the following:
- we create correct types for all fragments (like hash and dns)
- isValid automatically infer that the fragments are valid | skipped (not sure very useful)
- provide a helper to create types from the whole fragments but that keeps valid and skipped (basivally what the suggestion above does, but create directly a type, not infer it from an input)
- provide helpers to get valid / skipped / invalid / error for every fragment types (and eventual variation for v2/v3)
- make sure the lib is still extensible

### Initial message (for posterity)
I'm stuck => [Typescript Playground](https://www.typescriptlang.org/play?ssl=1&ssc=1&pln=15&pc=21#code/JYOwLgpgTgZghgYwgAgEJwM4QDwA0B8yA3sjAPZkBcyAFAJTIC8huAvgFCiSyIoBiFZBAAekEABMMaTDhABXALYAjaISIcu0eEmlQhoiBKnos2DGCigA5mo7sEZEOdIVqAsk2IuqtBs2QALBwOTmDISnBQ1Oh6jF7kPvRMhABEASl2Ic6RUHAAnp4A2gkANOGRALrs9o7OMCCeMHIgCGDAjnj4NBAANhAKhmDRMp2FFQxE7MjTyFlkfQB0PWRW3X0D4HRTM1AQYHJQDb39gwsKcAAO3clCCwn0W5m1YQiNIDQ5+XQA3EA)


cavacado:
**isValid automatically infer that the fragments are valid | skipped (not sure very useful)**
i'm assuming we will get this for free when the types of the fragments are well-defined?

do you suppose a map of fragments (you input these verifiers, you get these kinds of fragments) written in example form in some part of the documentation (or maybe as comments) would help in aiding the developer understand what kind of fragments are being returned? (this would be complementary to the types, instead of picturing the type of `OpenAttestationDnsTxtVerificationFragment` in one's mind, the developer would have a concrete object to inspect)

i'm all for creating better types for all the fragments, but to be honest i'm worried about whether future people are able internalise the types and not fight with the compiler.
(i guess that’s why you also suggested to have helpers to create the fragment type)

overall i think your suggestions are sound, but I'm just afraid that the type complexity might cause people to do castings instead of understanding the types.
but if its a one-off implementation then the rest of us would need to understand conditional types and `infer` more deeply (if we ever touch the types there).

`Promise<PromiseValue<ReturnType<T["verify"] | T["skip"]>>[]>` 🤯

src/getIdentifier.ts
```diff
@@ -1,80 +1,81 @@
...
-const getIdentityProofFragment = (fragments: VerificationFragment[]) => {
+const getIdentityProofFragment = <T extends AllVerificationFragment<unknown>>(fragments: T[]): T | undefined => {
   if (fragments.length < 1) {
     throw new Error("Please provide at least one verification fragment");
   }
   return fragments.find((status) => status.type === "ISSUER_IDENTITY" && status.status === "VALID");
 };
 
-export const getIdentifier = (fragments: VerificationFragment[]) => {
+export const getIdentifier = <T extends AllVerificationFragment<unknown>>(fragments: T[]) => {
```
yehjxraymond, line 73: I wonder for legacy reason do we want to combine the identifiers for all the `ISSUER_IDENTITY` fragments? If not its only returning the first which is pretty inaccurate if it's a group of issuers.

src/types/core.ts
```diff
@@ -43,40 +43,61 @@ export interface VerifierOptions {
...
+}
+export interface SkippedVerificationFragment extends VerificationFragment {
+  status: "SKIPPED";
+  reason: Reason;
+  // this is useless, it's just to make it easier to work with fragment, otherwise data is not available and need to narrow the type down first
+  data?: undefined;
 }
 export type VerificationFragmentType = "DOCUMENT_INTEGRITY" | "DOCUMENT_STATUS" | "ISSUER_IDENTITY";
 export type VerificationFragmentStatus = "ERROR" | "VALID" | "INVALID" | "SKIPPED";
+export type AllVerificationFragment<Data> =
```
yehjxraymond, line 74: As above, to use any since it's unlike any `Data` will be useful here.

src/types/core.ts
```diff
@@ -43,40 +43,61 @@ export interface VerifierOptions {
...
+  | ValidVerificationFragment<Data>
+  | InvalidVerificationFragment<Data>
+  | SkippedVerificationFragment
+  | ErrorVerificationFragment<Data>;
+
+export type VerificationFragmentWithData<Data> =
```
yehjxraymond, line 80: Similarly.

src/types/error.ts
```diff
@@ -92,8 +95,9 @@ export interface EthersError extends Error {
...
-  code: number;
-  codeString: string;
-  message: string;
-}
+export const Reason = Record({
+  code: Number,
+  codeString: String,
+  message: String,
+});
+export type Reason = Static<typeof Reason>;
```
yehjxraymond, line 103: Overloading here again

src/verifiers/documentStatus/didSigned/didSignedDocumentStatus.ts
```diff
@@ -227,8 +275,35 @@ const verifyV3 = async (
 
   const revokedOnAny = revocationStatus.revoked;
 
-  const status = issuedOnAll && !revokedOnAny;
+  if (ValidDidSignedIssuanceStatus.guard(verificationResult) && ValidRevocationStatus.guard(revocationStatus)) {
```
yehjxraymond, line 278: Quite weird to see so many guards here, was thinking if it makes sense to have one 

```
if(isValidStatus(verificationResult)){
  return valid stuffs
}else{
  return invalid stuffs (only one more guard is required here to discern between sign and revocation instead of the nested ternary?)
}
```

yehjxraymond: Or the pattern used in `src/verifiers/documentStatus/documentStore/ethereumDocumentStoreStatus.ts` for v2 is also pretty nice. 

if(revoked) return revoked
if(not signed) return not signed
return valid

src/verifiers/documentStatus/documentStore/ethereumDocumentStoreStatus.ts
```diff
@@ -198,46 +203,64 @@ const verifyV3 = async (
...
+    return {
+      name,
+      type,
+      data,
+      status: "VALID",
+    };
+  }
+
+  // eslint-disable-next-line no-nested-ternary
+  const reason = InvalidRevocationStatus.guard(revocation)
```
yehjxraymond, line 225: v2 flow up there seemed easier to read

src/verifiers/documentStatus/documentStore/ethereumDocumentStoreStatus.type.ts
```diff
@@ -0,0 +1,107 @@
...
+  RevocationStatus,
+  RevocationStatusArray,
+  ValidRevocationStatus,
+  ValidRevocationStatusArray,
+} from "../revocation.types";
+
+/**
+ * Document store issuance status
+ */
+export const ValidDocumentStoreIssuanceStatus = Record({
```
yehjxraymond, line 19: lots of overloading..

src/verifiers/documentStatus/revocation.types.ts
```diff
@@ -0,0 +1,22 @@
+import { Literal, Record, Static, String, Union, Array as RunTypesArray, Optional } from "runtypes";
+import { Reason } from "../../types/error";
+
+export const ValidRevocationStatus = Record({
```
yehjxraymond, line 4: overloading

src/verifiers/documentStatus/tokenRegistry/ethereumTokenRegistryStatus.type.ts
```diff
@@ -0,0 +1,79 @@
...
+import { Reason } from "../../../types/error";
+
+/**
+ * Token registry mint status
+ */
+export const ValidTokenRegistryStatus = Record({
+  minted: Literal(true),
+  address: String,
+});
+export type ValidTokenRegistryStatus = Static<typeof ValidTokenRegistryStatus>;
```
yehjxraymond, line 17: overloading
overloading

src/verifiers/issuerIdentity/did/didIdentityProof.ts
```diff
@@ -55,37 +61,59 @@ const verifyV2 = async (document: v2.WrappedDocument, options: VerifierOptions):
...
-    signatureVerifications.every((i) => i.status === "VALID");
+  const signatureVerifications: DidVerificationStatusArray = await Promise.all(signatureVerificationDeferred);
 
-  return {
-    name,
-    type,
-    data: signatureVerifications,
-    status: signedOnAll ? "VALID" : "INVALID",
-  };
+  if (signatureVerifications.length > 0 && ValidDidVerificationStatusArray.guard(signatureVerifications)) {
```
yehjxraymond, line 87: Rather having the length guard outside, will it make sense to have a Runtype guard check for the min length constraint in `ValidDidVerificationStatusArray` ?

src/verifiers/issuerIdentity/did/didIdentityProof.ts
```diff
@@ -55,37 +61,59 @@ const verifyV2 = async (document: v2.WrappedDocument, options: VerifierOptions):
...
+  if (InvalidDidVerificationStatus.guard(invalidSignature)) {
+    return {
+      name,
+      type,
+      data: signatureVerifications,
+      reason: invalidSignature.reason,
+      status: "INVALID",
+    };
+  }
+  throw new CodedError(
```
yehjxraymond, line 106: Will reach here if the array if there is no issuer in array, not sure if we want to catch that separately.

src/verifiers/issuerIdentity/dnsDid/dnsDidProof.ts
```diff
@@ -96,20 +94,32 @@ const verifyV2 = async (document: v2.WrappedDocument): Promise<VerificationFragm
     return verifyIssuerDnsDid({ key, location });
   });
   const verificationStatus = await Promise.all(deferredVerificationStatus);
-  const overallStatus =
-    verificationStatus.some((status) => status.status === "VALID") &&
-    verificationStatus.every((status) => status.status === "VALID")
-      ? "VALID"
-      : "INVALID";
+
+  if (verificationStatus.length > 0 && ValidDnsDidVerificationStatusArray.guard(verificationStatus)) {
```
yehjxraymond, line 98: same as above, might be better with length guard in runtype?

src/verifiers/issuerIdentity/dnsDid/dnsDidProof.type.ts
```diff
@@ -0,0 +1,55 @@
...
+
+/**
+ * DNS-DID verification status
+ */
+export const ValidDnsDidVerificationStatus = Record({
+  status: Literal("VALID"),
+  location: String,
+  key: String,
+});
+export type ValidDnsDidVerificationStatus = Static<typeof ValidDnsDidVerificationStatus>;
```
yehjxraymond, line 17: overload

src/verifiers/issuerIdentity/dnsTxt/openAttestationDnsTxt.ts
```diff
@@ -124,8 +122,17 @@ const verifyV2 = async (document: v2.WrappedDocument, options: VerifierOptions):
     })
   );
 
-  const invalidIdentity = identities.find((identity): identity is InvalidIdentity => identity.status !== "VALID");
-  if (invalidIdentity) {
+  if (identities.length > 0 && ValidDnsTxtVerificationStatusArray.guard(identities)) {
```
yehjxraymond, line 125: as above as runtype constraint

src/verifiers/issuerIdentity/dnsTxt/openAttestationDnsTxt.ts
```diff
@@ -156,42 +165,47 @@ const verifyV3 = async (document: v3.WrappedDocument, options: VerifierOptions):
...
+  }
   return {
     name,
     type,
     data: {
       identifier: issuerIdentity.location,
       value: issuerIdentity.value,
     },
-    ...(issuerIdentity.status === "INVALID" ? { reason: issuerIdentity.reason } : {}),
     status: issuerIdentity.status,
```
yehjxraymond, line 187: This is awkward now, when I highlight, ts tells me its "VALID", so might as well put it as valid?

yehjxraymond: Also, i think its better to guard the positive case and return invalid on else in case the guard fails to catch something?

src/verifiers/issuerIdentity/dnsTxt/openAttestationDnsTxt.type.ts
```diff
@@ -0,0 +1,77 @@
...
+
+/**
+ * DNS-TXT verification status
+ */
+export const ValidDnsTxtVerificationStatus = Record({
+  status: Literal("VALID"),
+  location: String,
+  value: String,
+});
+export type ValidDnsTxtVerificationStatus = Static<typeof ValidDnsTxtVerificationStatus>;
```
yehjxraymond, line 18: overload

README.md
```diff
@@ -182,6 +182,56 @@ isValid(fragments); // display false because ISSUER_IDENTITY is INVALID
 isValid(fragments, ["DOCUMENT_INTEGRITY", "DOCUMENT_STATUS"]); // display true because those types are VALID
 ```
 
+## Utils and types
```
yehjxraymond, line 185: might want to update readme on Usage section where it's showing the old fragment:

```
  {
    "name": "OpenAttestationDnsTxt",
    "type": "ISSUER_IDENTITY",
    "data": [
      {
        "status": "VALID",
        "location": "example.tradetrust.io",
        "value": "0xe59877ac86c0310e9ddaeb627f42fdee5f793fbe"
      }
    ],
    "status": "VALID"
  }
```

README.md
```diff
@@ -182,6 +182,56 @@ isValid(fragments); // display false because ISSUER_IDENTITY is INVALID
...
+    if (ValidTokenRegistryDataV2.guard(data)) {
+        // data is correctly typed here
+    }
+}
+```
+
+Note that in the example above, using `utils.isValidFragment` might be unnecessary. It's possible to use directly `ValidTokenRegistryDataV2.guard` over the data.
+
+### List of utilities
+- `getOpenAttestationHashFragment`
```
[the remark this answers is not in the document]
Nebulis: can add

src/common/utils.test.ts
```diff
@@ -0,0 +1,217 @@
...
+  getOpenAttestationDnsDidIdentityProofFragment,
+  getOpenAttestationDnsTxtIdentityProofFragment,
+  getOpenAttestationEthereumDocumentStoreStatusFragment,
+  getOpenAttestationEthereumTokenRegistryStatusFragment,
+  getOpenAttestationHashFragment,
+} from "./utils";
+import { AllVerificationFragment } from "..";
+
+describe("getFragments", () => {
+  const fragments: AllVerificationFragment<any>[] = [
```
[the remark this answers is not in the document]
Nebulis: will default to any

src/common/utils.test.ts
```diff
@@ -0,0 +1,217 @@
...
+      reason: {
+        code: 0,
+        codeString: "OOPS",
+        message: "Document was oopsed",
+      },
+      type: "ISSUER_IDENTITY",
+    },
+  ];
+  it("should return undefined when fragment does not exist", () => {
+    expect(getFragmentByName("NOPE")(fragments)).toMatchInlineSnapshot(`undefined`);
```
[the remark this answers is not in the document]
Nebulis: Then one that extends oa-verify with a custom verification method would not be able to use the function without a typescript error

src/common/utils.ts
```diff
@@ -9,3 +16,52 @@ export const getDefaultProvider = (options: VerificationBuilderOptionsWithNetwor
 export const getProvider = (options: VerificationBuilderOptions): providers.Provider => {
   return "provider" in options ? options.provider : getDefaultProvider(options);
 };
+
+export const getFragmentByName = <T extends VerificationFragment>(name: string) => (fragments: T[]): T | undefined =>
```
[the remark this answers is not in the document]
Nebulis: I'm not sure why you want curried function to have a specific name. Curried function are like normal function, but just call differently.

For instance lodash-fp does not rename the function name to something else.

If the meaning is ambiguous, I can change to "normal" function (no curry)

src/did/verifier.ts
```diff
@@ -1,21 +1,29 @@
...
 
-export interface ValidDidVerificationStatus {
-  verified: true;
-  did: string;
-}
+export const ValidDidVerificationStatus = Record({
+  verified: Literal(true),
+  did: String,
+});
+export type ValidDidVerificationStatus = Static<typeof ValidDidVerificationStatus>;
```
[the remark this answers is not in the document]
Nebulis: The document does the same => https://github.com/pelotom/runtypes#static-type-inference which I prefer over having Foo + FooRT types for each one.

Also, when :
- Foo is use as type, it's correctly inferred (no .guard, no .check, etc).
- Foo is used as a variable, it's correctly infered as well (.guard, .check, etc.)

src/getIdentifier.ts
Nebulis: I can open an issue to address this later

src/verifiers/documentStatus/didSigned/didSignedDocumentStatus.ts
Nebulis:
```ts
const issuance = await (await Promise.all(signatureVerificationDeferred)).map(transformToDidSignedIssuanceStatus);
  const notIssued = issuance.find(InvalidDidSignedIssuanceStatus.guard);
  const revoked = revocationStatuses.find(InvalidRevocationStatus.guard);

  const data = {
    issuedOnAll: !notIssued,
    revokedOnAny: !!revoked,
    details: {
      issuance,
      revocation: revocationStatuses,
    },
  };

  if (ValidDidSignedDataV2.guard(data)) {
    return {
      name,
      type,
      data,
      status: "VALID",
    };
  }

  let reason: Reason | undefined;

  if (InvalidDidSignedIssuanceStatus.guard(notIssued)) {
    reason = notIssued.reason;
  }
  if (InvalidRevocationStatus.guard(revoked)) {
    reason = revoked.reason;
  }
  if (!reason) {
    throw new CodedError(
      "Unable to retrieve the reason of the failure",
      OpenAttestationDidSignedDocumentStatusCode.UNEXPECTED_ERROR,
      "UNEXPECTED_ERROR"
    );
  }
  return {
    name,
    type,
    data,
    status: "INVALID",
    reason,
  };
```

Is this ok ?

- no more double guard
- the reason return will be issuance if issuance failed, or revocation, if revoked? (Is it what you meant by like document store ?)

src/verifiers/issuerIdentity/did/didIdentityProof.ts
Nebulis: yup, why not

src/verifiers/issuerIdentity/did/didIdentityProof.ts
Nebulis: technically it shouldn't, if there is no issuer, it will be skipped, unless calling verify straight (but that's not how the lib is used)

src/verifiers/documentStatus/didSigned/didSignedDocumentStatus.ts
yehjxraymond: lgtm!

src/verifiers/issuerIdentity/dnsTxt/openAttestationDnsTxt.ts
Nebulis: Can change to "VALID", I don't care much, same for the order

README.md
Nebulis: change what ?

src/did/verifier.ts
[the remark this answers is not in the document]
yehjxraymond: Kind of feel like the same argument as props:

```
export interface MyComponent {
  foo: Bar;
}

export const MyComponent: FunctionComponent<MyComponent> ... 
```

I recall you were taking the opposing stance even though the the type/var is correctly inferred

src/common/utils.ts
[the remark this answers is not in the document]
yehjxraymond: Yea that I agree, is just that the caller might not be expecting the function to be called twice?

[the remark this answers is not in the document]
yehjxraymond: A builder gets the function which is then used as the getter fn

src/common/utils.test.ts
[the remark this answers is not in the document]
yehjxraymond: Hmm good point

src/common/utils.ts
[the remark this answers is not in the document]
Nebulis: Well, a caller should read the function signature before using it :p I will chang to a normal function to avoid any ambiguity

README.md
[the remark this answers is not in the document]
Nebulis: Done, added `getDocumentIntegrityFragments`, `getDocumentStatusFragments`, and `getIssuerIdentityFragments`.

src/common/utils.test.ts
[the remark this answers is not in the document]
Nebulis: done

src/getIdentifier.ts
Nebulis: done

src/types/core.ts
Nebulis: done

src/types/core.ts
Nebulis: done

src/verifiers/documentStatus/didSigned/didSignedDocumentStatus.ts
Nebulis: done

src/verifiers/issuerIdentity/dnsTxt/openAttestationDnsTxt.ts
Nebulis: done

src/verifiers/documentStatus/documentStore/ethereumDocumentStoreStatus.ts
Nebulis: you prefer like this :
```ts
let reason: Reason | undefined;
  if (InvalidRevocationStatus.guard(revocation)) {
    reason = revocation.reason;
  } else if (InvalidDocumentStoreIssuanceStatus.guard(issuance)) {
    reason = issuance.reason;
  }
```

src/verifiers/issuerIdentity/did/didIdentityProof.ts
Nebulis: done

src/verifiers/issuerIdentity/dnsDid/dnsDidProof.ts
Nebulis: done

src/verifiers/issuerIdentity/dnsTxt/openAttestationDnsTxt.ts
Nebulis: done

README.md
Nebulis: done

src/types/core.ts
```diff
@@ -43,40 +43,64 @@ export interface VerifierOptions {
...
+export interface ErrorVerificationFragment<Data> extends VerificationFragment {
+  status: "ERROR";
+  reason: Reason;
+  data: Data;
+}
+export interface SkippedVerificationFragment extends VerificationFragment {
+  status: "SKIPPED";
+  reason: Reason;
+  // this is useless, it's just to make it easier to work with fragment, otherwise data is not available and a consumer needs to narrow the type down first
+  data?: undefined;
```
Nebulis, line 70: maybe never is better :)

yehjxraymond (approved):
lgtm

PR #170, opened 2021-03-26
feat: update to oa v5 for better type guards

Nebulis:
Updated to open-attestation v5, with better type guards. In this MR, I mainly updated the tests, so that eachverification method still uses the type guards provided.

However, there are some side effects. Some parts of the code are unreachable. Some verification methods that previously ran, are not skipped.

It won't change the overall behaviour (almost). I will add comments on parts of the code that must be discussed about


src/verifiers/documentIntegrity/hash/openAttestationHash.test.ts
```diff
@@ -134,7 +134,7 @@ describe("OpenAttestationHash", () => {
         },
       ]);
     });
-    it("should fail if the merkle root is missing", async () => {
+    it("should skip if the merkle root is missing", async () => {
```
Nebulis, line 137: Here, we don't run the verification method anymore, because the file is not considered valid.

src/verifiers/documentStatus/didSigned/didSignedDocumentStatus.test.ts
```diff
@@ -254,12 +254,12 @@ describe("verify", () => {
       const res = await openAttestationDidSignedDocumentStatus.verify(documentDidObfuscatedRevocation, options);
       expect(res).toMatchInlineSnapshot(`
         Object {
-          "data": [Error: revocation block not found for an issuer],
+          "data": [Error: Document does not match either v2 or v3 formats],
```
Nebulis, line 257: Same kind of issue here, however it should be skipped (I will fix after)

src/verifiers/documentStatus/didSigned/didSignedDocumentStatus.ts
```diff
@@ -52,6 +51,7 @@ const verifyV2 = async (
   const { issuers } = data;
 
   const revocation: (v2.Revocation | undefined)[] = issuers.map((issuer) => issuer.revocation);
+  // TODO unreachable
```
Nebulis, line 54: I added this kind of comments for parts of the code that should not be reachable anymore

src/verify.v2.integration.test.ts
```diff
@@ -700,23 +682,23 @@ describe("verify(integration)", () => {
     // Ethers would return INVALID_ARGUMENT, as merkle root is odd-length which we tampered it by removing the last char
     expect(isValid(results, ["DOCUMENT_INTEGRITY"])).toStrictEqual(false);
     expect(isValid(results, ["DOCUMENT_STATUS"])).toStrictEqual(false);
-    expect(isValid(results, ["ISSUER_IDENTITY"])).toStrictEqual(true);
+    expect(isValid(results, ["ISSUER_IDENTITY"])).toStrictEqual(false);
```
Nebulis, line 685: This test is probably the most interesting change that happens in the MR. The identity is not checked anymore, because the signature is not valid.

src/verify.v3.integration.test.ts
```diff
@@ -542,8 +542,8 @@ describe("verify v3(integration)", () => {
       expect(valid).toBe(true);
     });
     it("should return valid fragments for documents correctly issued but with data obfuscated", async () => {
-      const obfuscated = obfuscate(v3DidSigned, ["issuer"]);
-      expect(obfuscated.issuer).toBe(undefined);
+      const obfuscated = obfuscate(v3DidSigned, ["reference"]);
```
Nebulis, line 545: We can't obfuscate any data anymore for testing purposes. If the issuer is missing, the document will be considered invalid.

src/verify.v2.integration.test.ts
rjchow: should probably add this in the changelog

src/verifiers/documentStatus/didSigned/didSignedDocumentStatus.test.ts
```diff
@@ -319,7 +319,21 @@ describe("verify", () => {
...
+      const res = await openAttestationDidSignedDocumentStatus.verify(
+        {
+          ...documentDidSigned,
+          proof: [
+            {
+              created: "2021-03-25T07:52:31.291Z",
+              type: "OpenAttestationSignature2018",
+              proofPurpose: "assertionMethod",
+              verificationMethod: "did:ethr:0xdede",
+              signature: "0xabcf",
```
rjchow, line 331: i'm slightly surprised this doesn't fail for invalid length?

src/verifiers/documentStatus/didSigned/didSignedDocumentStatus.test.ts
```diff
@@ -624,12 +638,12 @@ describe("verify", () => {
...
-          "data": [Error: revocation type not found for an issuer],
+          "data": [Error: Document does not match either v2 or v3 formats],
           "name": "OpenAttestationDidSignedDocumentStatus",
           "reason": Object {
-            "code": 9,
-            "codeString": "UNRECOGNIZED_REVOCATION_TYPE",
-            "message": "revocation type not found for an issuer",
+            "code": 8,
+            "codeString": "UNRECOGNIZED_DOCUMENT",
+            "message": "Document does not match either v2 or v3 formats",
```
rjchow, line 646: feels like the error message became less useful

src/verifiers/documentStatus/didSigned/didSignedDocumentStatus.ts
rjchow: why TODO?

src/verifiers/documentStatus/didSigned/didSignedDocumentStatus.ts
```diff
@@ -79,6 +79,7 @@ const verifyV2 = async (
       case v2.RevocationType.None:
         return Promise.resolve({ revoked: false });
       default:
+        // TODO unreachable
```
rjchow, line 82: is this unreachable according to static verification or its actually runtime unreachable?

src/verifiers/documentStatus/tokenRegistry/ethereumTokenRegistryStatus.ts
```diff
@@ -173,12 +173,9 @@ const skip: VerifierType["skip"] = async () => {
 const test: VerifierType["test"] = (document) => {
   if (utils.isWrappedV2Document(document)) {
     const documentData = getData(document);
-    return !!documentData?.issuers?.some((issuer) => "tokenRegistry" in issuer);
-  }
-  if (utils.isWrappedV3Document(document)) {
-    const documentData = document?.openAttestationMetadata;
-    if (!!documentData?.proof && !!documentData?.identityProof && !!documentData?.template)
-      return document?.openAttestationMetadata?.proof?.method === v3.Method.TokenRegistry;
+    return documentData.issuers.some((issuer) => "tokenRegistry" in issuer);
```
rjchow, line 176: im too noob to understand this syntax

src/verify.v2.integration.test.ts
```diff
@@ -612,18 +612,18 @@ describe("verify(integration)", () => {
     expect(isValid(results)).toStrictEqual(true);
   });
   it("should be invalid with a merkle root that is odd-length", async () => {
+    // TODO some fragments are skipped because merkleRoot is weird. Do we want that to happen ? result is slightly different
```
rjchow, line 615: i think skipped is weird if the object key is correctly formed but the merkle root is corrupt

src/verify.v2.integration.test.ts
rjchow: related to above comment, dyt we should fix it so that if key correct but value malformed it should error? can be confusing to troubleshoot for new implementers if its just skipped and they dont know why

src/verify.v3.integration.test.ts
rjchow: ok noted

src/verifiers/documentStatus/didSigned/didSignedDocumentStatus.ts
Nebulis: wondering if we keep (in case) or remove

src/verifiers/documentStatus/didSigned/didSignedDocumentStatus.ts
Nebulis: runtime unreachable

src/verifiers/documentStatus/didSigned/didSignedDocumentStatus.test.ts
SiaChenRen: The version 2 document only check the required object in proof exist and the data type doesn't check the length.

src/verifiers/documentStatus/didSigned/didSignedDocumentStatus.ts
```diff
@@ -27,8 +27,7 @@ const skip: VerifierType["skip"] = async () => {
 const test: VerifierType["test"] = (document) => {
   if (utils.isWrappedV2Document(document) && utils.isSignedWrappedV2Document(document)) {
     return document.proof.some((proof) => proof.type === "OpenAttestationSignature2018");
-  }
-  if (utils.isWrappedV3Document(document) && utils.isSignedWrappedV3Document(document)) {
+  } else if (utils.isWrappedV3Document(document) && utils.isSignedWrappedV3Document(document)) {
```
SiaChenRen, line 30: i think we can remove the check for isWrappedV2Document and isWrappedV3Document since the check already exist in isSignedWrappedV2Document and isSignedWrappedV3Document

src/verifiers/issuerIdentity/did/didIdentityProof.ts
```diff
@@ -112,7 +111,7 @@ const verifyV3 = async (document: v3.WrappedDocument, options: VerifierOptions):
 const verify: VerifierType["verify"] = withCodedErrorHandler(
   async (document, options) => {
     if (utils.isWrappedV2Document(document)) return verifyV2(document, options);
-    if (utils.isWrappedV3Document(document)) return verifyV3(document, options);
+    else if (utils.isWrappedV3Document(document)) return verifyV3(document, options);
```
SiaChenRen, line 114: Just curious how come need to add the else since there is a return statement in the condition which act as a break before the next condition check?

src/verify.v2.integration.test.ts
Nebulis: fixed

src/verifiers/documentStatus/didSigned/didSignedDocumentStatus.ts
Nebulis: You are right, changed

src/verifiers/issuerIdentity/did/didIdentityProof.ts
Nebulis: imho it's more readable (can argue) but more importantly it's more logical., whether there is a return or not. I don't want to change the if/else structure depending on whether thee branch returns or not

src/verifiers/documentStatus/didSigned/didSignedDocumentStatus.test.ts
Nebulis: It's less useful I agree but fixing that would be a bit troublesome. Is there any huge concern ?

src/verifiers/documentStatus/didSigned/didSignedDocumentStatus.test.ts
Nebulis: There is no length check on the signature indeed

rjchow (approved):

PR #180, opened 2021-06-07
fix: add warning when configuration may lead to use of development ap…

Nebulis:
…i key


src/verifiers/verificationBuilder.ts
```diff
@@ -8,18 +8,26 @@ import {
...
  */
+let displayWarning = true;
 export const verificationBuilder = <T extends Verifier<any>>(
   verifiers: T[],
   builderOptions: VerificationBuilderOptions
 ) => (document: DocumentsToVerify, promisesCallback?: PromiseCallback): Promise<VerificationFragment[]> => {
+  // if the user didn't configure an API key and didn't configure a provider or a resolver, then he will likely use a development key. We then warn him once, that he may need to configure things properly, especially for production
+  if (displayWarning && (!builderOptions.resolver || !builderOptions.provider) && !process.env.INFURA_API_KEY) {
+    displayWarning = false;
+    console.warn(
```
rjchow, line 27: oa-verify doesn't have debug?

rjchow (approved):

src/verifiers/verificationBuilder.ts
```diff
@@ -8,18 +8,26 @@ import {
...
+      "You are using oa-verify default configuration, which is not suitable for production environment. Please make sure that yu configured the library correctly."
```
SiaChenRen, line 28: Have abit typo 
"Please make sure that yu configured the library correctly"
to 
"Please make sure that you configured the library correctly"

src/verifiers/verificationBuilder.ts
Nebulis: which means need to setup log level. Nah I want this one to be visible

src/verifiers/verificationBuilder.ts
Nebulis: thanks

src/verifiers/verificationBuilder.ts
rjchow: ok

PR #185, opened 2021-07-21
chroe: move snyk to devDep

eugbyte:
Right now, snyk is added as a direct dependency. This increases the size of the oa-verify library [by at least 20mb](https://www.npmjs.com/package/snyk).

I moved snyk dependency to a devDepency.

# To test
```
npm i
npm run snyk-protect
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,6 @@
     "ethr-did-resolver": "^4.3.3",
     "node-cache": "^5.1.2",
     "runtypes": "^6.3.0",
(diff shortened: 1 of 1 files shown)


Nebulis:
image

Looks like it's required in order to run snyk protect. Let's try to use tree shaking instead => https://webpack.js.org/guides/tree-shaking/

Please show the bundle before and after applying tree shaking https://www.npmjs.com/package/webpack-bundle-analyzer

PR #192, opened 2021-08-31
chore: provider test case in v2 and v3 integration

isaackps:
# In this PR: 
- remove extra condition check for provider in test cases in v2 and v3 integration test


src/verify.v2.integration.test.ts
```diff
@@ -1947,8 +1937,9 @@ describe("verify(integration)", () => {
   });
 
   it("should use the defaults to connect to provider even when process.env is not there for document with document store", async () => {
+    process.env.PROVIDER_NETWORK = "";
```
Nebulis, line 1940: I don't get the point of setting the value in the variable and using the env variable to build the option.

Why do we do so? Why don't we set directly the value inside the option:
```
network: ""
```

isaackps: because by right the function to get the provider should will use process.env.PROVIDER_NETWORK value, so this is to test the case where the process.env.PROVIDER_NETWORK is not present and that the function will fallback to use the default set in the function to generate the provider.

Nebulis: but if you use the same value for the rest and for the env variable, how do you know it will pickup what you expect ?

isaackps: erm good question, cos my verifier should be defaulted to mainnet, so if i verify a document with the mainnet verifier, it should be valid document, thats how i do the test in this case.

Nebulis: so the test is to verify on mainnet when there is nothing set?
- set nothing
- make it clearer in the test name

isaackps: ok i have been thinking about what u said and on second thought i think this test does not make sense, because there is a type check in verificationBuilder and i can't pass in empty network into the function. The test that test for the default values is already at the utils test file, so i'll be deleting this test case.

PR #193, opened 2021-08-31
chore: fix v3 test fixture that are not valid v3 credentials

Nebulis:


test/fixtures/v3/did-revocation-store-signed-not-revoked.json
```diff
@@ -12,13 +12,10 @@
   "validFrom": "2010-01-01T19:23:24Z",
   "issuer": {
     "id": "https://example.com",
+    "type": "OpenAttestationIssuer",
```
cavacado, line 15: i really need to read up more on json-ld, did you read the spec to understand the general concepts of it?

cavacado (commented):
this is because we forgot to add `location` into the schemata?
https://github.com/Open-Attestation/schemata/commit/a9f3a3a40f67db1137ac6559f8955103aea4f674#diff-b2db6399f89f73707a2a262cfbe6e66da50fe9c969e3eac7221983e923f87969R84

in reference to https://github.com/Open-Attestation/open-attestation/issues/201#issuecomment-906483500

test/fixtures/v3/did-revocation-store-signed-not-revoked.json
Nebulis: No it's because most of the files were not v3 valid due to the latest changes of context. In oa-verify we don't wrap document so we can't really see whether a file is valid or not

I added location for this MR because it was missing in the context.

I read the doc (not fully) but I didn't find it helpful. Fak3 helped me to understand some concepts (how to apply scope with types)

cavacado: ah ic.

👌

okok, then i'll read up the discussion you had with Fak3 that is somewhere around (plus the spec if im bored).

PR #196, opened 2021-09-21
chore: update-oa

eugbyte:
update open-attestation to latest version which uses  @govtechsg/jsonld


package.json
```diff
@@ -13,7 +13,7 @@
     "clean": "rm -rf dist/",
     "commit": "git-cz",
     "commit:retry": "npm run commit -- --retry",
-    "test": "jest",
+    "test": "jest util --silent=false",
```
Nebulis, line 16: nope

src/common/utils.test.ts
```diff
@@ -104,6 +104,13 @@ const fragments: AllVerificationFragment[] = [
 ];
 
 describe("getFragmentByName", () => {
+  it.only("should not flash warning", () => {
```
Nebulis, line 107: nope

src/common/utils.ts
```diff
@@ -55,6 +55,8 @@ export const generateProvider = (options?: ProviderDetails): providers.Provider
     );
   }
 
+  console.log("hello2");
```
Nebulis, line 58: nope

PR #205, opened 2021-10-28
feat: 🎸 added ocsp revocation for did documents

waynewee:
Added revocation of DID documents through an OCSP responder


src/types/core.ts
```diff
@@ -116,3 +120,11 @@ export interface ProviderDetails {
   url?: string;
   apiKey?: string;
 }
+
+export interface OcspResponse {
+  certId: string;
```
Nebulis, line 125: cert => certificate

Let's avoid shorter name for no reason

src/verifiers/documentStatus/utils.ts
```diff
@@ -63,6 +64,37 @@ export const isAnyHashRevoked = async (smartContract: DocumentStore, intermediat
...
+}: {
+  certId: string;
+  location: string;
+}): Promise<RevocationStatus> => {
+  const { data } = await axios.get(`${location}/${certId}`);
+
+  const {
+    certStatus,
+    reasonCode,
+  }: { certStatus: OcspResponderRevocationStatus; reasonCode: OcspResponderRevocationReason } = data;
```
Nebulis, line 79: let's use runtype to make sure the response format is correct

src/types/core.ts
```diff
@@ -116,3 +120,11 @@ export interface ProviderDetails {
   url?: string;
   apiKey?: string;
 }
+
+export interface OcspResponse {
```
Nebulis, line 124: what do you think to add a comment to the spec about ocsp response format ?

waynewee: yes will need to update that (you mean the adr right?)

Nebulis: I didn't mean the adr but I'm fine with a link to the adr too

src/verifiers/documentStatus/didSigned/didSignedDocumentStatus.test.ts
```diff
@@ -445,6 +453,113 @@ describe("verify", () => {
...
+            "details": Object {
+              "issuance": Array [
+                Object {
+                  "did": "did:ethr:0xE712878f6E8d5d4F9e87E10DA604F9cB564C9a89",
+                  "issued": true,
+                },
+              ],
+              "revocation": Array [
+                Object {
+                  "address": "https://www.ica.gov.sg/ocsp",
```
Nebulis, line 487: should it be https://www.ica.gov.sg/ocsp/SGCNM21566326 ?

src/types/core.ts
```diff
@@ -116,3 +120,11 @@ export interface ProviderDetails {
...
+  certificateId: string;
```
Nebulis, line 125: I think it's better to split the different kind of response (successful / error) rather than merging.

Eventually the error one can extend the successfull one.

wdyt?

src/verifiers/documentStatus/revocation.types.ts
```diff
@@ -20,3 +20,19 @@ export const RevocationStatus = Union(ValidRevocationStatus, InvalidRevocationSt
 export type RevocationStatus = Static<typeof RevocationStatus>;
 export const RevocationStatusArray = RunTypesArray(RevocationStatus);
 export type RevocationStatusArray = Static<typeof RevocationStatusArray>;
+
+export const OcspResponderRevocationStatus = Union(Literal("good"), Literal("revoked"), Literal("unknown"));
+export type OcspResponderRevocationStatus = Static<typeof OcspResponderRevocationStatus>;
+
+export enum OcspResponderRevocationReason {
```
Nebulis, line 27: Where does all this come from ?

test/fixtures/v2/did-revocation-ocsp-signed-revoked.json
```diff
@@ -0,0 +1,68 @@
+{
```
Nebulis, line 1: hummm actually this document is only revoked because of the mock right? Technically it's not revoked

waynewee: yeap, but nothing in the document tells you its revoked, the only thing that tells u its revoked is a CRL/OCSP

src/verifiers/documentStatus/revocation.types.ts
waynewee: for the runtype?

src/types/core.ts
waynewee: like a OcspResponseRevoked and OcspResponseNotRevoked interfaces?

Nebulis: yes

src/verifiers/documentStatus/revocation.types.ts
Nebulis: the different codes

test/fixtures/v2/did-revocation-ocsp-signed-revoked.json
Nebulis: it's more that if I see this document name, I might reused it in another test and expect it to be revoked (and fail)

any reason to use two different documents for the test? Can use the same document and configure msw to fail in one case and not the other right?

src/verifiers/documentStatus/revocation.types.ts
waynewee: https://datatracker.ietf.org/doc/html/rfc6960 page 6

test/fixtures/v2/did-revocation-ocsp-signed-revoked.json
waynewee: oh thats true actually, will change

src/verifiers/documentStatus/didSigned/didSignedDocumentStatus.test.ts
```diff
@@ -118,6 +123,7 @@ describe("verify", () => {
     it("should pass for documents using `DID` and is correctly signed", async () => {
```
Nebulis, line 123: missing tests for v3

package.json
```diff
@@ -13,7 +13,7 @@
     "clean": "rm -rf dist/",
     "commit": "git-cz",
     "commit:retry": "npm run commit -- --retry",
-    "test": "jest",
+    "test": "jest --runInBand",
```
Nebulis, line 16: nope, just for the ci =), so either add a new script or run `npm run test -- --runInBand` in circle ci

Nebulis (approved):

cavacado (approved):
i had a look at everything, got a little confused with the revoked reason codes as the spec points you to the CRL spec to get the number codes.
only got it resolved when i looked closely at the appendix section of the spec
```
RevokedInfo ::= SEQUENCE {
   revocationTime          GeneralizedTime,
   revocationReason    [0] EXPLICIT CRLReason OPTIONAL }
```
and
```
CRLReason ::= ENUMERATED {
        unspecified             (0),
        keyCompromise           (1),
        cACompromise            (2),
        affiliationChanged      (3),
        superseded              (4),
        cessationOfOperation    (5),
        certificateHold         (6),
             -- value 7 is not used
        removeFromCRL           (8),
        privilegeWithdrawn      (9),
        aACompromise           (10) }
```
i guess that is fine as you mentioned these in the ADR already.

apart from that i dont have any other comments

skydudie:
Hi @waynewee,

How would the OCSP revocation work for certificates that share the same certificate ID?
Currently larger institutions might not have unique certificate IDs across all their issuing registrars, as such revoking via certificate ID might cause the other certificates that share the same ID to be revoked as well. Is there a consideration to use something that will be unique across all certificates such as the targetHash of a document?

waynewee:
Hi @skydudie,

It is up to the implementers to decide on what exactly to use for the certificate ID. I believe the targetHash of a document is unique and you can use it as the certificate ID if desired.

skydudie:
Hi @waynewee thanks for the reply,

Just to clarify, that means for implementations where the certificate ID is not unique across they whole organisation, OCSP revocation would not work? Because OA-Verify will always take the certificate's 'id' as the parameter to check against the OSCP revocation list and the certificate id is not a unique value.

Besides, the targetHash is generated after the document is wrapped, so we won't be able to use it as the certificate ID.

waynewee:
Hi @skydudie,

You can give the certificate id any value. If the certificate ID is not unique across the whole organisation, then you have a number of options. One, you can setup two responders for each organisation in which certificate ids are always unique or two, you can generate new ids for all certificates and set that in the certificate id field.

PR #206, opened 2021-11-03
fix: move fragment error handling to utils instead

isaackps:
## Summary
Move fragments error handling into utils file


ChenRen93 (approved):
lgtm

PR #217, opened 2022-01-10
fix: verify issuer

simboonlong:
- reuse and export functions to verify issuer of various `IdentityProofType`
- use case = to validate txt records, to fail / disable UI early if not exists

say something like this for creator (not designed yet):

disabled


isaackps (approved):
LGTM! but anyone can eyeball the changes too!

simboonlong:
closing, using dnsprove directly instead, related pr https://github.com/TradeTrust/document-creator-website/pull/215

PR #232, opened 2022-07-27
fix: false positive for warnProvider message

marlowl:
As [suggested](https://docs.tradetrust.io/docs/appendix/tradetrust-api/#verify-endpoint), I was trying to setup my own version of the oa-verify api. I'm trying to target the Goerli test network, since Ropsten and Rinkeby will become deprecated in the future. (See [this](https://ethereum-magicians.org/t/og-council-post-merge-testnets/9034) for more info). 

I did come across the same issue as @HJunyuan described [here](https://github.com/Open-Attestation/oa-verify/issues/223). I looked at the logic of the code and I think just removing the warning message will fix the false positive. Feel free to correct/comment if removing this line has any other side effects for the package! 😄 

Fixes #223

Issue #223, opened 2022-03-24
False positive for warnProvider message


HJunyuan:
Even after specifying a custom provider and resolver, the `warnProvider` message still appears.

**Example code**:
```typescript
import { providers } from "ethers";
import { verificationBuilder, createResolver, openAttestationVerifiers } from "@govtechsg/oa-verify";

const NETWORK_NAME = "homestead";
const INFURA_API_KEY = "ENTER_KEY_HERE";

const verify = verificationBuilder(openAttestationVerifiers, {
  provider: new providers.InfuraProvider(NETWORK_NAME, INFURA_API_KEY),
  resolver: INFURA_API_KEY
    ? createResolver({
        ethrResolverConfig: {
          networks: [{ name: "mainnet", rpcUrl: `https://mainnet.infura.io/v3/${INFURA_API_KEY}` }],
        },
      })
    : undefined,
});
```

**Warning message triggered from**:
https://github.com/Open-Attestation/oa-verify/blob/b0fb5f482bb744e8ed07155fb6585f90f495520a/src/common/utils.ts#L66-L68

PR #233, opened 2022-07-28
fix: false positive warning in verificationBuilder

marlowl:
Related to this [PR](https://github.com/Open-Attestation/oa-verify/pull/232) and [issue](https://github.com/Open-Attestation/oa-verify/issues/223), was still receiving the `warnProvider` message, even though I configured a full custom provider.

I played around with the package and I think this small logic change will also fix this false positive warning for the verificationBuilder. It will still give the warning if you are using the default provider, but not when you configured your own provider or resolver now! 😄


simboonlong (approved):
🚀

PR #260, opened 2023-10-09
chore: expose verify signature

cavacado:
## Summary

What is the background of this pull request?
This pr exposes the `verifySignature` function.
This convenience function would be used for verification of did:eth documents without the use the hefty structure of a verifier.


cavacado:
closing since chore wouldnt trigger a release.

PR #273, opened 2023-12-21
feat: migrate to sepolia

zixiang2018:
## What this PR does
- migrate from Goerli to Sepolia testnet
- refactor "goerli" text to "sepolia"
- update [test cases in utils.test.ts](https://github.com/Open-Attestation/oa-verify/pull/273/files#diff-b8da6918cd3b457997415e333281b3f816b1731a45ae13ec580f0722d0ac2457R307) to check that provider does not have api key, because [we are using a workaround to directly call the Alchemy URL (i.e. we use JsonRpcProvider for Sepolia Alchemy Provider case). ](https://github.com/Open-Attestation/oa-verify/blob/3553d97919692728e132183d070bbad302c68c5c/src/common/utils.ts#L75)
    - This testcase will fail once we bump `ethers` to v6 and use the sepolia-supported AlchemyProvider, which consist of `apiKey` property. 
    - Hence, I added a `TODO` tag so that we will remember to revert the change in the future bump
 - [modified the resolver config for this testcase](https://github.com/Open-Attestation/oa-verify/pull/273/files#diff-ae7d0293dd0fa71af42952c3207ec38d615e369017742e85940862cc2c38899aR101-R105) because Sepolia is not supported in our current version of `ethr-did-resolver` (we are using the pre-ethers v6 version and v6 has support for Sepolia but is a breaking change)


src/did/resolver.test.ts
```diff
@@ -128,16 +138,16 @@ describe("getProviderConfig", () => {
 
   it("should use env variables for provider config properties for didResolver", () => {
     process.env.PROVIDER_ENDPOINT_TYPE = "alchemy";
-    process.env.PROVIDER_NETWORK = "goerli";
+    process.env.PROVIDER_NETWORK = "sepolia";
 
     expect(getProviderConfig()).toEqual({
-      networks: [{ name: "goerli", rpcUrl: "https://eth-goerli.alchemyapi.io/v2/_gg7wSSi0KMBsdKnGVfHDueq6xMB9EkC" }],
+      networks: [{ name: "sepolia", rpcUrl: "https://eth-sepolia.g.alchemy.com/v2/" }],
```
zixiang2018, line 144: No default API key used here because we are using a workaround with manual URL (`JsonRPCProvider`) for ethers v5 to call Sepolia. Default API key will apply in ethers v6 when we use `AlchemyProvider` which supports Sepolia

HJunyuan (approved):
🚀

PR #306, opened 2024-11-25
feat: add support for ocsp reason code suspended

chershentan:
## Summary
To support the [new reason code](https://github.com/Open-Attestation/ocsp-responder/pull/21) `1001` which represents the `suspended` document status on [OCSP responder](https://github.com/Open-Attestation/ocsp-responder)

## Changes

- update `ValidOcspReasonCode` to validate reason code `1001`
- update `OcspResponderRevocationReason` to return `SUSPENDED` for reason code `1001`
- Add test cases for these updates

### DIFF
--- a/src/verifiers/documentStatus/didSigned/didSignedDocumentStatus.type.ts
+++ b/src/verifiers/documentStatus/didSigned/didSignedDocumentStatus.type.ts
@@ -7,6 +7,7 @@ import {
 } from "../../../types/core";
 import { Reason } from "../../../types/error";
 import {
+  OcspResponderRevocationReason,
   RevocationStatus,
   RevocationStatusArray,
   ValidRevocationStatus,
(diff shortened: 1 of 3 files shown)


src/verifiers/documentStatus/didSigned/didSignedDocumentStatus.type.ts
```diff
@@ -40,7 +40,7 @@ export type DidSignedIssuanceStatusArray = Static<typeof DidSignedIssuanceStatus
  * OCSP response
  */
 
-export const ValidOcspReasonCode = Number.withConstraint((n) => n >= 0 && n <= 10 && n != 7);
+export const ValidOcspReasonCode = Number.withConstraint((n) => [1, 2, 3, 4, 5, 6, 8, 9, 10, 1001].includes(n));
```
superical, line 43: I was just thinking, would this be simpler and still work?
```
export const ValidOcspReasonCode = Number.withConstraint((n) => Object.values(OcspResponderRevocationReason).includes(n));
```

This will need to import `OcspResponderRevocationReason` from `revocation.types.ts` into this file though. But what you have works too!

superical: Another thing is, `n >= 0 && n <= 10 && n != 7` includes `0` but `[1, 2, 3, 4, 5, 6, 8, 9, 10, 1001]` doesn't. So maybe need to add `0` if we prefer listing out the numbers.

chershentan: Good catch!
I will follow your suggestion as it is cleaner

superical (approved):

